PR #23, opened 2021-05-31
Fixed account data

### LINKED ISSUE #20: Simplify AccountInfo
AccountInfo Data field cannot be deduced. Revert to returning the encoded string + encoding

tiago18c:
| Ready | Bug & Refactor | Yes | Closes #20  |

## Problem

Account info method and subscription would cause exceptions in the json parsin.

## Solution

Its impossible to know what the object we're parsing during the getAccountInfo method, so the idea is to return the base64 encoded data and let the user figure out what to do with it.


murlokito (approved):
LGTM

Issue #15, opened 2021-05-30
Add tests for Solnet.Rpc


murlokito:
Add tests to the RPC project attempting to cover the most important methods:

- [x] GetAccountInfo
- [x] GetBalance
- [x] GetRecentBlockhash
- [x] GetSupply
- [x] GetTokenAccountBalance
- [x] GetTokenAccountsByDelegate
- [x] GetTokenAccountsByOwner
- [x] GetTokenSupply
- [x] GetTokenLargestAccounts

PR #25, opened 2021-06-01
Add tests for the RPC client

murlokito:
| Ready | Testing | No | #15  |

## Problem

Lack of tests in the http rpc client

## Solution

Added tests for all implemented methods.


src/Solnet.Rpc/Models/Logs.cs
```diff
@@ -34,4 +36,38 @@ public class LogInfo : Log
...
+    public class InstructionError
+    {
+        public int ErrorCode { get; set; }
+        
+        public KeyValuePair<string, int> CustomError { get; set; }
+    }
+    
+    public class InstructionErrors
+    {
+        [JsonPropertyName("InstructionError")]
```
tiago18c, line 49: Redundant. Doesn't diverge from the property name

src/Solnet.Rpc/Models/Logs.cs
```diff
@@ -34,4 +36,38 @@ public class LogInfo : Log
...
+        [JsonConverter(typeof(InstructionErrorJsonConverter))]
+        public InstructionError[] InstructionError { get; set; }
```
tiago18c, line 51: I was thinking more along the lines of a list<object> or object[]. Are we sure that there are no more types of errors other than these two fields within InstructionError ?

test/Solnet.Rpc.Test/UtilitiesTest.cs
```diff
@@ -0,0 +1,66 @@
...
+        public void TestCreateProgramAddressException()
+        {
+            _ = AddressExtensions.CreateProgramAddress(
+                new[] { Encoding.UTF8.GetBytes("SeedPubey1111111111111111111111111111111111") },
+                Encoding.UTF8.GetBytes(LoaderProgramId));
+        }
+        
+        /*
+        [TestMethod]
+        public void TestCreateProgramAddress()
```
tiago18c, line 26: Why not delete instead?

tiago18c (commented):
Please review comments.

src/Solnet.Rpc/Models/Logs.cs
murlokito: Not because of the serializer options

src/Solnet.Rpc/Models/Logs.cs
murlokito: No but it can be refactored later, so far I've only seen these types when attempting transactions with possible errors

tiago18c: I guess we could create an issue for a later milestone (0.2 or 1.0?) to review the rust code and try to find out what we can actually find out in these errors?!

src/Solnet.Rpc/Utilities/AddressExtensions.cs
```diff
@@ -18,23 +20,23 @@ public static class AddressExtensions
         /// <exception cref="Exception">Throws exception when the resulting address doesn't fall off the Ed25519 curve.</exception>
         public static byte[] CreateProgramAddress(IList<byte[]> seeds, byte[] programId)
         {
-            var buffer = new List<byte>();
+            var buffer = new MemoryStream();
```
tiago18c, line 23: Initialize the MemoryStream with (32*seed + "ProgramDerivedAddress".Length + programId.Length) as initial capacity.

murlokito: This is done

tiago18c (approved):
Send it!

Issue #27, opened 2021-06-02
Add logger to clients and implement client factory with client interfaces


murlokito:
Add a way to inject a logger that implements `Microsoft.Extensions.Logging.ILogger` and implement a client factory as well as interfaces to not expose the client's constructors

---
- [x] RPC Client Logging
- [ ] Streaming RPC Client Logging
---
- [x] RPC Client Interface
- [x] Streaming RPC Client Interface
---
- [x] Client Factory

PR #28, opened 2021-06-02
Add logging to rpc client, implement factory and interfaces

murlokito:
| Ready | Feature/Tooling/Refactor | Yes | #27 |

## Problem

Not exposing library internals like:
- the injection of the http client which is used for testing only;
- adding logging capabilities;
- implementing an interface and having the ClientFactory return that interface instead of the actual class that implements it

## Solution

Implemented:
- proper logging on the rpc client (streaming rpc is still to do)
- interfaces and clientfactory ready for both clients


tiago18c (approved):
Pump it.

Issue #31, opened 2021-06-02
Full JSON RPC API coverage


murlokito:
Provide an implementation for each of the following HTTP JSON RPC API methods:

- [X] getAccountInfo
- [X] getBalance
- [x] getBlock
- [X] getBlockHeight
- [x] getBlockProduction
- [X] getBlockCommitment
- [x] getBlocks
- [x] getBlocksWithLimit
- [X] getBlockTime
- [X] getClusterNodes
- [x] getEpochInfo
- [x] getEpochSchedule
- [x] getFeeCalculatorForBlockhash
- [x] getFeeRateGovernor
- [x] getFees
- [x] getFirstAvailableBlock
- [X] getGenesisHash
- [x] getHealth
- [X] getIdentity
- [X] getInflationGovernor
- [X] getInflationRate
- [X] getInflationReward
- [X] getLargestAccounts
- [x] getLeaderSchedule
- [x] getMaxRetransmitSlot
- [x] getMaxShredInsertSlot
- [X] getMinimumBalanceForRentExemption
- [x] getMultipleAccounts
- [x] getProgramAccounts
- [X] getRecentBlockhash
- [x] getRecentPerformanceSamples
- [x] getSignaturesForAddress
- [x] getSignatureStatuses
- [x] getSlot
- [x] getSlotLeader
- [x] getSlotLeaders
- [x] getStakeActivation
- [X] getSupply
- [X] getTokenAccountBalance
- [X] getTokenAccountsByDelegate
- [X] getTokenAccountsByOwner
- [X] getTokenLargestAccounts
- [X] getTokenSupply
- [x] getTransaction
- [X] getTransactionCount
- [x] getVersion
- [x] getVoteAccounts
- [X] minimumLedgerSlot
- [X] requestAirdrop
- [X] sendTransaction
- [X] simulateTransaction

Provide an implementation for each of the following WS JSON RPC API methods:

- [X] accountSubscribe
- [X] accountUnsubscribe
- [X] logsSubscribe
- [X] logsUnsubscribe
- [X] programSubscribe
- [X] programUnsubscribe
- [X] signatureSubscribe
- [X] signatureUnsubscribe
- [X] slotSubscribe
- [X] slotUnsubscribe

Mark them as done as we go, once all are done, close the issue and move on, cool guys don't look at explosions.

tiago18c:
Done :)

PR #54, opened 2021-06-04
Implements all Slot related methods and fixing some docs

### LINKED ISSUE #53: Implement Slot related rpc methods
Implement all `Slot` related rpc methods

murlokito:
| Ready | Feature/Bug/Refactor/ | No | #31 Closes #53 |

## Problem

getMinimumBalanceForRentExemption is untested
no implementation for the following `Slot`-related methods

- getSlot
- getSlotLeaders
- getSlotLeader
- getSnapshotSlot
- getMaxRetransmitSlot
- getMaxShredInsertSlot

as well as `getLargestAccounts`

## Solution

Added test for getMinimumBalanceForRentExemption
Implemented the `Slot` methods and while fixing docs also added the `LargestAccounts` method in order to solve naming convention conflict with the new `LargeTokenAccounts`


src/Solnet.Rpc/Models/Transaction.cs
```diff
@@ -0,0 +1,7 @@
+namespace Solnet.Rpc.Models
+{
+    public class Transaction
+    {
+
```
tiago18c, line 5: Why is this empty?

tiago18c (commented):
Just wondering about that empty Transaction class. If you delete it consider this approved.

src/Solnet.Rpc/Models/Transaction.cs
murlokito: Had already written that before we figured your structures for blocks would be useful for getTransaction and you would get it done in the same PR. Removed.

PR #85, opened 2021-06-11
Fix cake build order

### LINKED ISSUE #83: Fix the cake build order
As it stands the cake build is skipping steps and jumping from the `Test` task straight to the `Pack` task, this should be fixed, `Pack` should be dependent on `Publish`, so it only happens when everything is OK.

murlokito:
| Ready | Fix | No | Closes #83  |

## Problem

Fixing the cake build order due to missing files in the code coverage report

## Solution

Changed the `Pack` task to only run after `Publish`

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -7,6 +7,12 @@
       "commands": [
         "dotnet-cake"
       ]
+    },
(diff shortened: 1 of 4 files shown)


tiago18c (approved):
LGTM

PR #87, opened 2021-06-11
Improving keystore to allow restoring wallet from string byte array

### LINKED ISSUE #82: Improve the KeyStore
Improve the KeyStore to allow recovering wallets from a string which represents the private key's byte array, as follows:

`[69,191,12,22,125,16,119,72,240,150,74,197,249,221,54,164,172,222,248,202,22,242,96,43,105,164,101,52,155,41,46,6,107,27,120,68,31,183,113,110,148,151,206,38,195,198,108,78,97,66,196,191,82,41,240,33,253,9,89,19,75,196,171,104]`

murlokito:
| Ready | Refactor | Yes | Closes #82 |

## Problem

The keystore compatible with the `solana-keygen` tool only allows to restore the wallet from the file.

## Solution

Changed the keystore to allow restoring from both a file and a string that represents the same contents that the file would have,
so users can import keys via applications (e.g. Sollet using import private key)

### DIFF
--- a/src/Solnet.KeyStore/SolanaKeyStore.cs
+++ b/src/Solnet.KeyStore/SolanaKeyStore.cs
@@ -9,30 +9,25 @@ namespace Solnet.KeyStore
     /// </summary>
     public class SolanaKeyStoreService
     {
+        /// <summary>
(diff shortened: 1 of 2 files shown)


tiago18c (approved):
LGTM

Issue #94, opened 2021-06-16
Remove NBitcoin dependency


murlokito:
Remove NBitcoin dependency by integrating the mnemonic and seed generation into  `Solnet.Wallet`.

PR #95, opened 2021-06-16
Remove NBitcoin dependency

### LINKED ISSUE #90: Slash dependencies
After analyzing how long it takes to load a Blazor Wasm application we realized we can do a better job at keeping the library lightweight so the idea is to slash some of the following dependencies:

- [ ] NBitcoin is only used because of the Mnemonic, we could easily add only the needed code to the `Solnet.Wallet` project and save around ~4MB of total size.

- [x] BouncyCastle is referenced in the `Solnet.Wallet` project but isn't used there, only in the KeyStore.

murlokito:
| Ready | Refactor | Yes | Closes #90 #94  |

## Problem

- Dependency slashing to improve DLL load times in Blazor Wasm applications etc

## Solution

- Integrates the necessary features from NBitcoin in order to provide mnemonic and seed generation to be used with Ed25519


src/Solnet.Wallet/Bip39/KdTable.cs
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        private static bool Supported(char c)
+        {
+            return _SupportedChars.Any(r => r[0] <= c && c <= r[1]);
+        }
+
+
+        static int[][] _SupportedChars = new int[][]{
+new[]{0,1000},
```
tiago18c, line 60: is this properly formatted?

src/Solnet.Wallet/Bip39/KdTable.cs
```diff
@@ -0,0 +1,76 @@
...
+
+using System;
+using System.Linq;
+using System.Text;
+
+namespace Solnet.Wallet.Bip39
+{
+    class KDTable
+    {
+        public static string NormalizeKD(string str)
```
tiago18c, line 11: make it internal instead

src/Solnet.Wallet/Bip39/Language.cs
```diff
@@ -0,0 +1,47 @@
+// unset
+
+namespace Solnet.Wallet.Bip39
+{
+    /// <summary>
+    /// Specifies the available languages for mnemonic generation.
+    /// </summary>
+    public enum Language
```
tiago18c, line 8: This should also be internal

src/Solnet.Wallet/Bip39/Mnemonic.cs
```diff
@@ -0,0 +1,231 @@
...
+
+namespace Solnet.Wallet.Bip39
+{
+    /// <summary>
+    /// A .NET implementation of the Bitcoin Improvement Proposal - 39 (BIP39)
+    /// BIP39 specification used as reference located here: https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki
+    /// This implementation was taken from NBitcoin: https://github.com/MetacoSA/NBitcoin/
+    /// </summary>
+    public class Mnemonic
+    {
```
tiago18c, line 14: if the Mnemonic is supposed to be used only by the wallet, make it internal.

src/Solnet.Wallet/Bip39/Wordlist.cs
```diff
@@ -0,0 +1,457 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace Solnet.Wallet.Bip39
+{
+    public class Wordlist
+    {
```
tiago18c, line 12: Same comment about class visibility.

tiago18c (commented):
Instead of copy&pasting more comments, Im going to make a general remark:

Make sure that all classes added from NBitcoin are internal, unless you think there is a chance they might be usefull for outside usage. Right now it seems to be very mixed, with some private, some internal, and a lot of public.

Also, all public methods and classes should be documented to avoid more warnings during compilation and pipeline runs.

tiago18c (approved):
LFG

PR #127, opened 2021-06-27
Make JsonRpcBase an abstract base type

skynode:
Since `JsonRpcBase` is intended to be a base class, make it impossible for the type to be instantiated anywhere else.


murlokito:
Makes perfect sense to me. Going to approve and after @tiago18c approval will merge! Thanks for your contribution :)

murlokito (approved):

tiago18c (approved):
Looks good. Thanks for the contribution 😁

PR #138, opened 2021-07-02
Enhancements and utilities

### LINKED ISSUE #135: More QoL updates
The following QoL and _possibly_ performance improvements should be implemented.

- Add functionality for Bit Flags as they exist in Serum and Mango to Solnet.Programs
- Add extension methods for `ReadOnlySpan<byte>` in order to make program data deserialization less verbose
- Refactor Solnet.Programs to use those extension methods where possible
- Migrate from the utilitites currently being used to encode U32LE and U64LE towards `BinaryPrimitives`

murlokito:
| Ready | Feature/Tooling/Refactor | No | Closes #135  |

## Problem

Making program data encoding and decoding less verbose
Adding utility classes that can be reused across several programs which use bitmasks to indicate account types

## Solution

Added two extension method classes, one for decoding and another for encoding
Added an abstract generic class `Flag` which can easily be subclassed to add functionality for a certain mask
Added `ByteFlag` class which subclasses `Flag` for masks with byte values


src/Solnet.Programs/NameServiceProgram.cs
```diff
@@ -228,15 +226,15 @@ public static PublicKey DeriveNameAccountKey(byte[] hashedName, PublicKey nameCl
...
 
-            methodBuffer[0] = (byte)NameServiceInstructions.Create;
-            Utils.Uint32ToByteArrayLe((ulong) hashedName.Length, methodBuffer, 1);
-            Array.Copy(hashedName, 0, methodBuffer, 5, hashedName.Length);
-            Utils.Int64ToByteArrayLe(lamports, methodBuffer, 37);
-            Utils.Uint32ToByteArrayLe((ulong) space, methodBuffer, 45);
+            methodBuffer.WriteU8((byte)NameServiceInstructions.Create, 0);
+            methodBuffer.WriteU32((uint) hashedName.Length, 1);
+            methodBuffer.WriteSpan(hashedName, 5);
+            methodBuffer.WriteU64(lamports, 37);
```
tiago18c, line 236: I have to say that its so much better to look at and reason with after these changes

tiago18c (approved):
Just Send It^tm

PR #146, opened 2021-07-11
Change PublicKeys constant length access modifier to be public (#145)

### LINKED ISSUE #145: Change PublicKey.PublicKeyLength to use "public" modifier
**Is your feature request related to a problem? Please describe.**
I'd like to have a public, hardcoded `int` of PublicKey length (always `32`). It is currently marked as `private`. This will be used for Offset layout addition, instead of a local variable in each deserialization class, or `32` defined in an unnecessary, duplicate place.

**Describe the solution you'd like**
Switch access modifier for PublicKey.PublicKeyLength from `private` to `public`.

**Describe alternatives you've considered**
Hard-code `32` somewhere else, but this is duplicate code and harder to maintain.

skynetcapital:
| Ready | Refacto | No | Closes #145  |

## Problem

I'd like to have a public, hardcoded int of PublicKey length (always 32). It is currently marked as private. This will be used for Offset layout addition, instead of a local variable in each deserialization class, or 32 defined in an unnecessary, duplicate place.

## Solution

Switched access modifier for PublicKey.PublicKeyLength from private to public.

### DIFF
--- a/SharedBuildProperties.props
+++ b/SharedBuildProperties.props
@@ -2,7 +2,7 @@
   xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <Product>Solnet</Product>
-    <Version>0.3.2</Version>
+    <Version>0.3.3</Version>
     <Copyright>Copyright 2021 &#169; Solnet</Copyright>
     <Authors>blockmountain</Authors>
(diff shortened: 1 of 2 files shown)


murlokito (approved):
Approved, LGTM :)

PR #154, opened 2021-07-19
Refactored AccountMetas, Programs implementations, Builders and interop features.

### LINKED ISSUE #152: [Bug] AccountMeta does not match what it should be
**Describe the bug**
The code interface is awkward or incorrect, when an account need to be a signer, `Account` is provided
https://github.com/bmresearch/Solnet/blob/93b568458abe5f866d0e030ac0fe1369f32e2d03/src/Solnet.Rpc/Models/AccountMeta.cs#L41-L48

I think that is quite a big change that makes it misaligned to how the rust solana_sdk or web3.js works. This also destroys the opportunity to build the transaction with C# then provide it to another component to be signed, as any signer need to be provided at instruction construction. So impossible to achieve multisig from different physical places, signing by hardware wallet...

It is quite important to respect the same interface between abstractions to also avoid a steeper learning curve between each abstraction.

I propose to just do it like web3.js
https://solana-labs.github.io/solana-web3.js/modules.html#AccountMeta
Another option is to follow the rust way, with a static method that sets one of the properties `AccountMeta.readOnly(publicKey, isSigner)`

Then providing Signer[] is done at the transaction level

### LINKED ISSUE #153: Refactor AccountMetas, TransactionBuilder, MessageBuilder
⚠️ BREAKING CHANGES ⚠️  

Related to #152 

In order to achieve a more modular approach and allow more interoperability with other ecosystem SDKs and even more use cases the following critical components of the SDK should be refactored and with certain features being added.

Refactor:
- `AccountMeta` previously had a single constructor that received the `PublicKey` instead of having two which could receive an `Account`, the goal is to go backwards to the old approach but adding a static method `ReadOnly(bool isSigner)` to have it similar to the Rust SDK
- `TransactionBuilder` currently is able to build the transaction with the necessary signatures because the `Account`s are passed directly into the `TransactionInstruction`s and are stored to be used during signing, this should also be reverted in order to allow the previous refactoring to happen, in the end, the old `Build(IList<Account> singers)` we had is what we should once again have

Features:
- Make the `Message` class public (see [Message in web3.js](https://solana-labs.github.io/solana-web3.js/classes/Message.html))
- Add a `Transaction` class which is similar to the one in web3.js as well [see](https://solana-labs.github.io/solana-web3.js/classes/Transaction.html)

With the additions mentioned above, we can achieve full interoperability with web3.js, rust sdk, hardware components etc.

Suggestions, comments etc are appreciated and encouraged.

murlokito:
| Hold | Feature/Refactor/Hotfix | Yes | Closes #152 #153  |

## Problem

Described in #152 and #153

### DIFF
--- a/src/Solnet.Examples/TransactionBuilderExample.cs
+++ b/src/Solnet.Examples/TransactionBuilderExample.cs
@@ -82,11 +82,11 @@ namespace Solnet.Examples
             Console.WriteLine($"MinBalanceForRentExemption Mint Account >> {minBalanceForExemptionMint}");
             
             var mintAccount = wallet.GetAccount(94206);
-            Console.WriteLine($"MintAccount: {mintAccount.PublicKey.Key}");
+            Console.WriteLine($"MintAccount: {mintAccount}");
             var ownerAccount = wallet.GetAccount(10);
-            Console.WriteLine($"OwnerAccount: {ownerAccount.PublicKey.Key}");
+            Console.WriteLine($"OwnerAccount: {ownerAccount}");
             var initialAccount = wallet.GetAccount(64209);
-            Console.WriteLine($"InitialAccount: {initialAccount.PublicKey.Key}");
+            Console.WriteLine($"InitialAccount: {initialAccount}");
 
             var tx = new TransactionBuilder().SetRecentBlockHash(blockHash.Result.Value.Blockhash)
                 .SetFeePayer(ownerAccount)
@@ -147,13 +147,13 @@ namespace Solnet.Examples
             Console.WriteLine($"MinBalanceForRentExemption Account >> {minBalanceForExemptionAcc}");
 
             var mintAccount = wallet.GetAccount(31);
-            Console.WriteLine($"MintAccount: {mintAccount.PublicKey.Key}");
+            Console.WriteLine($"MintAccount: {mintAccount}");
             var ownerAccount = wallet.GetAccount(10);
-            Console.WriteLine($"OwnerAccount: {ownerAccount.PublicKey.Key}");
+            Console.WriteLine($"OwnerAccount: {ownerAccount}");
             var initialAccount = wallet.GetAccount(32);
-            Console.WriteLine($"InitialAccount: {initialAccount.PublicKey.Key}");
+            Console.WriteLine($"InitialAccount: {initialAccount}");
             var newAccount = wallet.GetAccount(33);
-            Console.WriteLine($"NewAccount: {newAccount.PublicKey.Key}");
+            Console.WriteLine($"NewAccount: {newAccount}");
 
             var tx = new TransactionBuilder().SetRecentBlockHash(blockHash.Result.Value.Blockhash)
                 .SetFeePayer(ownerAccount)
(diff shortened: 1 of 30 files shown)


src/Solnet.Rpc/Builders/MessageBuilder.cs
```diff
@@ -96,12 +61,17 @@ internal int Length()
         /// <summary>
         /// The list of instructions contained within this transaction.
         /// </summary>
-        internal readonly List<TransactionInstruction> Instructions;
+        internal List<TransactionInstruction> Instructions { get; private set; }
```
tiago18c, line 64: its already internal, so I think the private set should either apply to all members, or none.
As long as the API consumer can't access this and screw up, should be good, but need consistency tho'

src/Solnet.Rpc/Builders/TransactionBuilder.cs
```diff
@@ -53,13 +54,14 @@ private IEnumerable<Account> GetSigners()
             
             foreach (TransactionInstruction instruction in _messageBuilder.Instructions)
             {
-                signers.AddRange(from accountMeta in instruction.Keys where accountMeta.Signer && !signers.Contains(accountMeta.Account) select accountMeta.Account);
+                signers.AddRange(from accountMeta in instruction.Keys where accountMeta.IsSigner && !signers.Contains(accountMeta.Account) select accountMeta.Account);
```
tiago18c, line 57: commented code => delete

tiago18c: Reason: we don't need commented code for history purposes when the code is already versioned.

tiago18c (commented):
Left some comments :)

src/Solnet.Examples/TransactionBuilderExample.cs
```diff
@@ -198,41 +204,87 @@ public void Run()
             Console.WriteLine($"MinBalanceForRentExemption Account >> {minBalanceForExemptionAcc}");
 
             var mintAccount = wallet.GetAccount(21);
-            Console.WriteLine($"MintAccount: {mintAccount.PublicKey}");
+            Console.WriteLine($"MintAccount: {mintAccount.PublicKey.Key}");
```
tiago18c, line 207: Override the ToString method to return the Key property. It will make these debug messages less verbose

src/Solnet.Examples/TransactionDecodingExample.cs
```diff
@@ -0,0 +1,124 @@
...
+                    minBalanceForExemptionAcc,
+                    TokenProgram.TokenAccountDataSize,
+                    TokenProgram.ProgramIdKey)).AddInstruction(TokenProgram.InitializeAccount(
+                    initialAccount.PublicKey,
+                    mintAccount.PublicKey,
+                    ownerAccount.PublicKey)).AddInstruction(TokenProgram.MintTo(
+                    mintAccount.PublicKey,
+                    initialAccount.PublicKey,
+                    1_000_000,
+                    ownerAccount)).AddInstruction(MemoProgram.NewMemo(initialAccount, "Hello from Sol.Net"))
```
tiago18c, line 70: Fix identation.

src/Solnet.Programs/TokenProgram.cs
```diff
@@ -36,14 +36,14 @@ public static class TokenProgram
...
+        /// <param name="owner">The public key of the owner account.</param>
         /// <returns>The transaction instruction.</returns>
         public static TransactionInstruction Transfer(
-            PublicKey source, PublicKey destination, ulong amount, Account owner)
+            PublicKey source, PublicKey destination, ulong amount, PublicKey owner)
         {
             List<AccountMeta> keys = new()
             {
-                new AccountMeta(source, true), new AccountMeta(destination, true), new AccountMeta(owner, false)
+                AccountMeta.Writable(source, false), AccountMeta.Writable(destination, false), AccountMeta.ReadOnly(owner, true)
```
tiago18c, line 46: Make indentation consistent with remaining instructions.

tiago18c (approved):
It looks good. I don't really see anything to pick apart other than indentations. LFG ◎

PR #155, opened 2021-07-21
Added GetConfirmedTransaction/Async methods as they are missing for m…

aph5nt:
…ainnet beta usage

| Ready | Feature | No | N/A |

## Problem

Mainnet-beta still runs solana-core v1.6 and the methods that got added on v1.7 to replace those that will be deprecated after v1.8 aren't available

## Solution

Implemented `GetConfirmedTransaction` as `Obsolete` to use it on mainnet because `GetTransaction` is only available on solana-core v1.7+

### DIFF
--- a/src/Solnet.Rpc/IRpcClient.cs
+++ b/src/Solnet.Rpc/IRpcClient.cs
@@ -796,7 +796,7 @@ namespace Solnet.Rpc
         /// <summary>
         /// Returns transaction details for a confirmed transaction.
         /// </summary>
-        /// <param name="signature">ransaction signature as base-58 encoded string.</param>
+        /// <param name="signature">Transaction signature as base-58 encoded string.</param>
         /// <param name="commitment">The state commitment to consider when querying the ledger state.</param>
         /// <returns>Returns a task that holds the asynchronous operation result and state.</returns>
         Task<RequestResult<TransactionMetaSlotInfo>> GetTransactionAsync(string signature,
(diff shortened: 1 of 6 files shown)


src/Solnet.Rpc/IRpcClient.cs
```diff
@@ -810,11 +817,20 @@ public interface IRpcClient
...
+        /// <returns>Returns an object that wraps the result along with possible errors with the request.</returns>
+        [Obsolete]
+        RequestResult<TransactionMetaSlotInfo> GetConfirmedTransaction(string signature);
+        
         /// <summary>
         /// Gets the total transaction count of the ledger.
         /// </summary>
         /// <param name="commitment">The state commitment to consider when querying the ledger state.</param>
         /// <returns>Returns a task that holds the asynchronous operation result and state.</returns>
+        [Obsolete]
```
murlokito, line 833: Looks like you marked `GetTransactionCountAsync` as Obsolete when it isn't? At least from what I'm seeing in the docs it's not

murlokito (changes requested):
Could you add a test for `GetConfirmedTransaction`? Should be similar to the existing test for `GetTransaction` [here](https://github.com/bmresearch/Solnet/blob/master/test/Solnet.Rpc.Test/SolanaRpcClientBlockTests.cs#L274)

src/Solnet.Rpc/IRpcClient.cs
aph5nt: you are right

aph5nt: fixed

murlokito: Looks like it's still there, you might have forgotten to add the changes to that file in the most recent commit, just fix that and I'll get it merged :)

murlokito (approved):
Thanks for this addition! Will merge once checks pass :)

PR #157, opened 2021-07-22
Change TokenBalance Amount type from string to ulong

### LINKED ISSUE #156: Change TokenBalance Amount type from string to ulong
Is it a good idea to change public string Amount { get; set; } into ulong (there is UiAmountString already if someone will need string)? Now I have to parse a string into ulong in my code. 

    /// <summary>
    /// Represents the token balance of an account.
    /// </summary>
    public class TokenBalance
    {
        /// <summary>
        /// The raw token account balance without decimals.
        /// </summary>
        public string Amount { get; set; }

        /// <summary>
        /// The number of base 10 digits to the right of the decimal place.
        /// </summary>
        public int Decimals { get; set; }

        /// <summary>
        /// The token account balance, using mint-prescribed decimals. DEPRECATED.
        /// </summary>
        [Obsolete("UiAmount is deprecated, please use UiAmountString instead.")]
        public decimal? UiAmount { get; set; }

        /// <summary>
        /// The token account balance as a string, using mint-prescribed decimals.
        /// </summary>
        public string UiAmountString { get; set; }
    }

aph5nt:


aph5nt:
closed

murlokito (approved):
Looks good to me!

PR #160, opened 2021-07-23
Adds ATA example and fixes a bug in ATA program and PDAs.

### LINKED ISSUE #158: Example of sending TOKEN to external account would be helpful.
Hi guys, 
I'm trying to send the existing token to an external account. So far I have had no issues with transferring it to myself, or to an account that is aware already of the given token. 

The problem is that I don't know how to Initialize an external account, please advise. You can find below the code I was playing with.

``` 
var tx = new TransactionBuilder().
                SetRecentBlockHash(blockHash.Result.Value.Blockhash).
                SetFeePayer(ownerAccount).
                AddInstruction(SystemProgram.CreateAccount(
                    ownerAccount,
                    new Account(string.Empty, message.Address),  // <== here is a problem, I don't know the private key of external address
                    minBalanceForExemptionAcc,
                    SystemProgram.AccountDataSize,
                    TokenProgram.ProgramIdKey)).
                AddInstruction(TokenProgram.InitializeAccount(toTokenAccountPublicKey, new PublicKey(tokenContractAddress), toTokenAccountPublicKey)).
                AddInstruction(TokenProgram.Transfer(
                    fromTokenAccountPublicKey,
                    toTokenAccountPublicKey,
                    lamportsAmount,
                    ownerAccount))
                .Build();
```

### LINKED ISSUE #159: Add example usage of Associated Token Accounts
- [ ] Add an example that uses ATAs to transfer tokens between accounts.

- [ ] Add a public method to `AssociatedTokenAccountProgram` to derive the ATA for a certain `owner` and `mint`.

- [ ] Fix a bug in `FindProgramAddress` where once the while loop began the nonce would be 254.

Related to #158

murlokito:
| Ready | Feature/Tooling/Refactor/Hotfix | No | Closes #159 Closes #158 |

## Problem

There was no example of using ATAs to send tokens to
While implementing an example I found out that some things weren't optimal for the library usage, such as:
 - `AssociatedTokenAccountProgram`  didn't have a public method to derive the ATA address
 - The `CreateAssociatedTokenAccount` instruction had a bug in the ATA key being non-writable
 - There was also an obscure bug in the `FindProgramAddresss` function where the nonce would start at 254 instead of 255, making most transactions using PDAs broken 

## Solution

- Added an example that creates, initializes and mints tokens to an account, polls GetTransaction until it's Finalized and then sends another transaction with a `CreateAssociatedTokenAccount` instruction and a `TokenProgram.Transfer` (ATA throws errors when the TX where the mint account is created and initialized is not Finalized)
- Fixed the `CreateAssociatedTokenAccount` instruction bug where the ATA key was non-writable
- Fixed the  `FindProgramAddresss` bug so it properly starts at 255
- Also fixed an issue in the `AccountKeysList` so it doesn't bork when a key is already present but we're trying to add it again as being writable on the transaction.

### DIFF
--- a/src/Solnet.Examples/AssociatedTokenAccountsExample.cs
+++ b/src/Solnet.Examples/AssociatedTokenAccountsExample.cs
@@ -1,0 +1,157 @@
+using Solnet.Programs;
+using Solnet.Rpc;
+using Solnet.Rpc.Builders;
+using Solnet.Rpc.Core.Http;
+using Solnet.Rpc.Messages;
+using Solnet.Rpc.Models;
+using Solnet.Rpc.Types;
+using Solnet.Wallet;
+using System;
+using System.Threading;
+
+namespace Solnet.Examples
+{
+    public class AssociatedTokenAccountsExample : IExample
+    {
+        
+        private static readonly IRpcClient RpcClient = ClientFactory.GetClient(Cluster.TestNet);
+
+        private const string MnemonicWords =
+            "route clerk disease box emerge airport loud waste attitude film army tray " +
+            "forward deal onion eight catalog surface unit card window walnut wealth medal";
+
+        public void Run()
+        {
+            Wallet.Wallet wallet = new Wallet.Wallet(MnemonicWords);
+            
+            /*
+             * The following region creates and initializes a mint account, it also creates a token account
+             * that is initialized with the same mint account and then mints tokens to this newly created token account.
+             */
+            #region Create and Initialize a token Mint Account
+            
+            
+            RequestResult<ResponseValue<BlockHash>> blockHash = RpcClient.GetRecentBlockHash();
+            
+            ulong minBalanceForExemptionAcc =
+                RpcClient.GetMinimumBalanceForRentExemption(SystemProgram.AccountDataSize).Result;
+            ulong minBalanceForExemptionMint =
+                RpcClient.GetMinimumBalanceForRentExemption(TokenProgram.MintAccountDataSize).Result;
+            
+            Console.WriteLine($"MinBalanceForRentExemption Account >> {minBalanceForExemptionAcc}");
+            Console.WriteLine($"MinBalanceForRentExemption Mint Account >> {minBalanceForExemptionMint}");
+
+            Account ownerAccount = wallet.GetAccount(10);
+            Account mintAccount = wallet.GetAccount(1002);
+            Account initialAccount = wallet.GetAccount(1102);
+            Console.WriteLine($"OwnerAccount: {ownerAccount.PublicKey.Key}");
+            Console.WriteLine($"MintAccount: {mintAccount.PublicKey.Key}");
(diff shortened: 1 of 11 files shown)


murlokito:
@aph5nt This PR fixes bugs related to your issue and adds an example, thanks for bringing this to my attention because really there were some broken things that needed fixing :)

src/Solnet.Examples/AssociatedTokenAccountsExample.cs
```diff
@@ -0,0 +1,157 @@
...
+                RpcClient.GetMinimumBalanceForRentExemption(TokenProgram.TokenAccountDataSize).Result;
+            ulong minBalanceForExemptionMint =
+                RpcClient.GetMinimumBalanceForRentExemption(TokenProgram.MintAccountDataSize).Result;
+            
+            Console.WriteLine($"MinBalanceForRentExemption Account >> {minBalanceForExemptionAcc}");
+            Console.WriteLine($"MinBalanceForRentExemption Mint Account >> {minBalanceForExemptionMint}");
+
+            Account ownerAccount = wallet.GetAccount(10);
+            Account mintAccount = wallet.GetAccount(1002);
+            Account initialAccount = wallet.GetAccount(1102);
```
skynetcapital, line 46: I don't know what these magic numbers mean. (Although this is just a test)

src/Solnet.Rpc/Utilities/AddressExtensions.cs
```diff
@@ -75,6 +78,7 @@ public static (byte[] Address, int Nonce) FindProgramAddress(IEnumerable<byte[]>
                 catch (Exception)
                 {
                     buffer.RemoveAt(buffer.Count - 1);
+                    nonce--;
```
skynetcapital, line 81: Is exception-driven logic the only solution here?

skynetcapital (commented):
Left 2 minor comments. Otherwise LGTM

src/Solnet.Examples/AssociatedTokenAccountsExample.cs
murlokito: These numbers are the index for the account, i.e. When you get a mnemonic on Sollet and use "Add Account" in the menu it's equivalent to using an index like that and incrementing it for everytime you go on the menu to add a new account without importing by private key.

src/Solnet.Rpc/Utilities/AddressExtensions.cs
murlokito: Probably not, would require refactoring CreateProgramAddress and FindProgramAddress to return a `null` or something when it can't find an address that doesn't fall off the curve

tiago18c: I would prefer the TrySomething approach
```
bool TryCreateProgramAddress(buffer, programId, out result);
```

murlokito: Going with TrySomething approach

PR #163, opened 2021-07-24
Implements the remaining methods that are deprecated on v1.7 

### LINKED ISSUE #162: Implement the remaining methods that are deprecated on v1.7 and removed on v1.8
The rationale behind this:

- It is needed to create explorers (and many other things!) using Solnet because mainnet-beta still uses v1.6
- As it stands, in order to properly query transaction and block data on mainnet you need to run your own node using v1.7
- In the end, it allows us to _actually_ have full JSON RPC API coverage, and it can be removed once v1.8+ rolls onto mainnet

murlokito:
| Ready | Feature | No | Closes #162  |

## Problem

See #162

### DIFF
--- a/src/Solnet.Rpc/SolanaRpcClient.cs
+++ b/src/Solnet.Rpc/SolanaRpcClient.cs
@@ -73,20 +73,20 @@ namespace Solnet.Rpc
         }
 
         private KeyValue HandleCommitment(Commitment parameter, Commitment defaultValue = Commitment.Finalized)
-        {
-            if (parameter != defaultValue)
-            {
-                return KeyValue.Create("commitment", parameter);
-            }
-            return null;
-        }
+         =>  parameter != defaultValue ? KeyValue.Create("commitment", parameter) : null;
+        
+        private KeyValue HandleTransactionDetails(TransactionDetailsFilterType parameter, 
+            TransactionDetailsFilterType defaultValue = TransactionDetailsFilterType.Full)
+            => parameter != defaultValue ? KeyValue.Create("transactionDetails", parameter) : null;
+
         #endregion
 
 
-
         #region Accounts
+
         /// <inheritdoc cref="IRpcClient.GetAccountInfoAsync(string,Commitment)"/>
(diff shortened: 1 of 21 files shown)


src/Solnet.Rpc/SolanaRpcClient.cs
```diff
@@ -149,25 +160,34 @@ public RequestResult<ResponseValue<AccountInfo>> GetAccountInfo(string pubKey, C
...
             if (commitment == Commitment.Processed)
             {
                 throw new ArgumentException("Commitment.Processed is not supported for this method.");
             }
 
             return await SendRequestAsync<BlockInfo>("getBlock",
-                Parameters.Create(slot, ConfigObject.Create(KeyValue.Create("encoding", "json"), HandleCommitment(commitment))));
+                Parameters.Create(slot, ConfigObject.Create(
+                    KeyValue.Create("encoding", "json"), HandleTransactionDetails(transactionDetails),
+                    blockRewards ? KeyValue.Create("rewards", true) : null, HandleCommitment(commitment))));
```
tiago18c, line 183: manually separate each parameter per line for better visibility

tiago18c: for readability, do the ternary as the Create parameter like in all other cases
```
KeyValue.Create("rewards", blockRewards ? true : null)
```

src/Solnet.Rpc/SolanaRpcClient.cs
```diff
@@ -180,18 +200,82 @@ public RequestResult<BlockInfo> GetBlock(ulong slot, Commitment commitment = Com
...
+            if (commitment == Commitment.Processed)
+            {
+                throw new ArgumentException("Commitment.Processed is not supported for this method.");
+            }
+
+            return await SendRequestAsync<BlockInfo>("getConfirmedBlock",
+                Parameters.Create(slot, ConfigObject.Create(
+                    KeyValue.Create("encoding", "json"),
+                    HandleTransactionDetails(transactionDetails),
+                    blockRewards ? KeyValue.Create("rewards", true) : null, HandleCommitment(commitment))));
```
tiago18c, line 222: same here about the ternary

tiago18c (approved):
Looks good. Merge after taking care of comments.

PR #165, opened 2021-07-26
Changed delegatedAmount from ulong to TokenBalance object as it should be

### LINKED ISSUE #164: [Bug] GetTokenAccountsByOwner deserialization error with delegatedAmount account edge case
**Describe the bug**
Found a JSON deserialization issue with the rpcClient.GetTokenAccountsByOwner method...
By fluke, a token account has the following data in the parsed->info response... 

                                 "delegatedAmount": {
                                    "amount": "45",
                                    "decimals": 9,
                                    "uiAmount": 4.5e-8,
                                    "uiAmountString": "0.000000045"
                                }

Most accounts usually have:

                                 "delegatedAmount": 12345

The TokenAccountInfoDetails class has DeletedAmount declared as an ulong, so the above structured data is making the serializer go pop.

    public ulong DelegatedAmount { get; set; }

The error is reported as the generic "Unable to parse json.", but the deserializer's innards say: 
    The JSON value could not be converted to System.UInt64. Path: $.result.value[42].account.data.parsed.info.delegatedAmount 

**To Reproduce**

var token_accounts = client.GetTokenAccountsByOwner(account, tokenProgramId:="TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA");

Shared the account that fails - don't want to share the pubkey in this issue.

**Expected behavior**
Deserialization should succeed :-)

**Additional context**
Looks as though the account in question may be related to staking in solfarm.

murlokito:
| Ready | Bug/Hotfix | Yes | Closes #164 |

## Problem

See #164 

## Solution

Changed  `DelegatedAmount` from `ulong` to `TokenBalance` as it should be

### DIFF
--- a/src/Solnet.Examples/GetTokenAccountsByOwnerExample.cs
+++ b/src/Solnet.Examples/GetTokenAccountsByOwnerExample.cs
@@ -1,0 +1,58 @@
+using Solnet.Rpc;
+using Solnet.Rpc.Core.Http;
+using Solnet.Rpc.Messages;
+using Solnet.Rpc.Models;
+using Solnet.Wallet;
+using System;
+using System.Collections.Generic;
+
+namespace Solnet.Examples
+{
+    public class GetTokenAccountsByOwnerExample : IExample
+    {
+        private static readonly IRpcClient rpcClient = ClientFactory.GetClient(Cluster.TestNet);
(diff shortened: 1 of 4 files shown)


tiago18c (approved):
LGTM;

PR #167, opened 2021-07-26
Changed getProgramAccounts memcmp filter to be a list

### LINKED ISSUE #166: Change `getProgramAccounts` memcmp filter to be a list
The `filters` in the `getProgramAccounts` method are an array and thus can have several `memcmp` filters (which is useful for Serum).

Create a model for the `MemCmp` and change the parameter to take a `List<MemCmp>`, internally traverse that list and add those to the request paremeters.

murlokito:
| Ready | Feature/Refactor | Yes | Closes #166 |

## Problem

See #166

The `filters` are a list and can have several `memcmp` filters (comes in handy for Serum OOAs for example)

## Solution

Added a MemCmp model
Refactored the `getProgramAccounts` signature to take a `List<MemCmp>` instead of just a single offset and bytes.

### DIFF
--- a/src/Solnet.Rpc/SolanaRpcClient.cs
+++ b/src/Solnet.Rpc/SolanaRpcClient.cs
@@ -104,31 +104,30 @@ namespace Solnet.Rpc
 
         /// <inheritdoc cref="IRpcClient.GetProgramAccountsAsync"/>
         public async Task<RequestResult<List<AccountKeyPair>>> GetProgramAccountsAsync(string pubKey,
-            Commitment commitment = Commitment.Finalized, int? dataSize = null, int? memOffset = null,
-            string memBytes = null)
+            Commitment commitment = Commitment.Finalized, int? dataSize = null, IList<MemCmp> memCmpList = null)
         {
-            Dictionary<string, object> memCmp = memOffset != null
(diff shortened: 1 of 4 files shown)


skynetcapital (approved):
LGTM

src/Solnet.Rpc/SolanaRpcClient.cs
```diff
@@ -104,31 +104,30 @@ private KeyValue HandleCommitment(Commitment parameter, Commitment defaultValue
...
-            Dictionary<string, object> memCmp = memOffset != null
-                ? ConfigObject.Create(
-                    KeyValue.Create("offset", memOffset),
-                    KeyValue.Create("bytes", memBytes))
-                : null;
+            Commitment commitment = Commitment.Finalized, int? dataSize = null, IList<MemCmp> memCmpList = null)
+        {
+            List<object> filters = Parameters.Create(ConfigObject.Create(KeyValue.Create("dataSize", dataSize)));
+            if (memCmpList != null)
+            {
```
tiago18c, line 111: Since you're already doing an ```if ``` , move the initial ```Parameters.Create ``` inside it.

tiago18c (approved):
LGTM;

Issue #168, opened 2021-07-27
Allow access to raw request/response JSON via RequestResult<T>


liquizard:
**Is your feature request related to a problem? Please describe.**
This feature request will extend the utility of the solnet lib beyond the intended happy path.
In unexpected situations where JSON serialization fails - rather than leave the dev/user at a dead end, access to the underlying JSON would enable creative workarounds or perhaps more information to be able to resolve or better diagnose an obstacle.
Another benefit would be greater transparency of the RPC chatter back and forth. Whilst this may not be perfectly aligned with the goals of the solnet project - I think some increase in the visibility of what is going on under the hood is beneficial to all parties in a number of different circumstances. Mostly the powerful magic of .Net JSON deserialization is wonderful until you encounter a breakage, and then it can be difficult to troubleshoot - especially in the hidden depths of a third party code.

**Describe the solution you'd like**
Extending the solnet RequestResult<T> class to provide:

        public string RawJsonRequest { get; set; }
        public string RawJsonResponse { get; set; }

**Describe alternatives you've considered**
Lots of diagnosis is already available via the extensive logging already implemented - but this is not optimal for programmatic inspection or when debugging. I think that adding specific fields offers better semantics and creates more opportunities for a developer to investigate and create solutions to their own problems.

Another alternative would be to throw a custom "JSON unmarshalling exception" that includes the req/response bodies as properties of the exception. This would serve in circumstances where there are mismatches between the expected and actual JSON schemas, but would not allow a curious developer from seeing the payloads of healthy RPC interactions.

**Additional context**
Here a screenshot of the properties at work - I found this useful in understanding and investigating #164  

Also the "unable to parse json" error message is slightly misleading as it sent me down the rabbit hole of thinking the JSON was invalid (badly escaped or truncated buffer etc. ) - the JSON was legit, but the problem was it did not match the expected class shape for deserialization. Perhaps this message could be something like: unable to deserialize json to {typename}

![image](https://user-images.githubusercontent.com/87999215/127123866-9fe94d20-778d-44af-a9d2-57508f806857.png)

Happy to submit a PR if you like the idea(s).

murlokito:
As we discussed following up with your report or #164 I really like this proposal, happy to review a PR with these additions anytime 😀

murlokito:
Closed by #169

PR #169, opened 2021-07-27
Create access methods to inspect RPC req/resp JSON on RequestResult<T>

liquizard:
## Problem
Please see #168 - Allow access to raw request/response JSON via RequestResult<T>

## Solution
Added RawRequest/RawResponse to JsonRpcRequest to share outside of scope to support end goal of passing on to  RequestResult<T>
Changed the name from the suggested RawJsonResponse to RawRpcResponse as its conceivable response may not be JSON at all (e.g. HTML proxy error etc)

## Other changes (e.g. bug fixes, small refactors)
Used internal setters so contents are immutable to callers outside of the solnet library.
Dropped requestJson local variable in JsonRpcClient.Task in favor of reusing newly added JsonRpcRequest.RawRequest

### DIFF
--- a/src/Solnet.Rpc/Core/Http/JsonRpcClient.cs
+++ b/src/Solnet.Rpc/Core/Http/JsonRpcClient.cs
@@ -61,32 +61,33 @@ namespace Solnet.Rpc.Core.Http
         /// <returns>A task that represents the asynchronous operation that holds the request result.</returns>
         protected async Task<RequestResult<T>> SendRequest<T>(JsonRpcRequest req)
         {
-            RequestResult<T> result;
-            req.RawRequest = JsonSerializer.Serialize(req, _serializerOptions);
+            var requestJson = JsonSerializer.Serialize(req, _serializerOptions);
 
             try
             {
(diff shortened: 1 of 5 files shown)


murlokito (approved):
This looks very good to me! But I think @tiago18c had a slightly different idea for this so I will let him take a look as well :)

tiago18c (changes requested):
I forgot that we had changed the serialization/desserialization earlier to/from a string every time, instead of just doing it conditionally depending on the log level.

Right now, I like it, just gonna nitpick about something, it feels dirty to hack the values into the JsonRpcRequest object.
I think I would prefer if the SendRequest was changed from
```
return await HandleResult<T>(req, response).ConfigureAwait(false);
```
into something like
```
result = await HandleResult<T>(req, response).ConfigureAwait(false);
result.RawRequest = blabla;
return result;
```
With these changes, we wouldn't need to store anything in the JsonRpcRequest object

liquizard:
Done. I've dropped req/resp from JsonRpcRequest and reworked using you suggest.
Much tidier - good shout.

tiago18c (approved):
Looks perfect now. LGTM;

PR #170, opened 2021-07-27
Implement InstructionDecoder (#161)

### LINKED ISSUE #161: Implement a POC for generic TransactionInstruction decoding
Request for comment on a "generic" instruction decoder, rough idea follows.

- Programs implement an interface which allows to decode some arbitrary data into a friendly instruction structure
- `Solnet.Programs` has a "`InstructionDecoder`" where you can register different programs which implement that interface
- You can call a method in the `InstructionDecoder` which attempts to decode the instruction

skynetcapital:
## Problem

Need to implement an `InstructionDecoder` as requested in (Closes #161)

## Solution

Implemented an `InstructionDecoder` static class.

### DIFF
--- a/src/Solnet.Examples/InstructionDecoderExample.cs
+++ b/src/Solnet.Examples/InstructionDecoderExample.cs
@@ -1,0 +1,20 @@
+﻿// unset
+
+using Solnet.Programs;
+using System;
+using System.Text;
+
+namespace Solnet.Examples
+{
+    public class InstructionDecoderExample : IExample
+    {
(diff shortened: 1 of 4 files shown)


murlokito (approved):
This looks like a good POC to me and I'm interested in playing around with it along with the code I'm preparing for #153 ahead of v0.4, can we get this merged? @tiago18c

tiago18c (approved):
It looks good to me. Just needs a little bit of housekeeping that can be done later with more decoding implementations.

PR #173, opened 2021-08-06
Implement Instruction Decoders

### LINKED ISSUE #172: Implement Instruction Decoders for other programs
Following up on #161, Instruction Decoders should be implemented for the following programs:

- TokenProgram
- SystemProgram
- ATAProgram
- NameServiceProgram
- SharedMemoryProgram

murlokito:
| Ready | Feature/Refactor/Hotfix | Yes | Closes #172  |

## Problem

See #172 

## Solution

Implemented instruction decoders for all implemented programs
Added more tests to cover some features which were added to Solnet.Wallet (Account, PK, PubKey operators, ToString etc)


test/Solnet.Programs.Test/InstructionDecoderTest.cs
```diff
@@ -0,0 +1,631 @@
...
+            Assert.AreEqual("System Program", decodedInstructions[0].ProgramName);
+            Assert.AreEqual("11111111111111111111111111111111", decodedInstructions[0].PublicKey);
+            Assert.AreEqual(0, decodedInstructions[0].InnerInstructions.Count);
+            Assert.IsTrue(decodedInstructions[0].Values.TryGetValue("From Account", out object fromAccount));
+            Assert.IsTrue(decodedInstructions[0].Values.TryGetValue("To Account", out object toAccount));
+            Assert.IsTrue(decodedInstructions[0].Values.TryGetValue("Base Account", out object baseAccount));
+            Assert.IsTrue(decodedInstructions[0].Values.TryGetValue("Seed", out object seed));
+            Assert.IsTrue(decodedInstructions[0].Values.TryGetValue("Space", out object space));
+            Assert.IsTrue(decodedInstructions[0].Values.TryGetValue("Amount", out object amount));
+            Assert.IsTrue(decodedInstructions[0].Values.TryGetValue("Owner", out object ownerAccount));
```
tiago18c, line 492: Comment for future reference:
In cases like this, were you have multiple steps before the last assert, you don't need to assert every step of the way if its easily traceable the origin of the problem.
For instance, in this case, you could've missed the `IsTrue` asserts for the `TryGetValue`, because for each of the `AreEqual` bellow to be true, its guaranteed that the `TryGetValue` had to succeed.

Another example would be
```
Assert.NotNull(foo);
Assert.IsTrue(foo.Bar);
```

If `foo` was ever null, you'd find out right in the next assert with a NPE, so the `Assert.NotNull` becomes redundant, and kinda pollutes the unit test. 

Not an issue tho, just saved keystrokes in the future :)

test/Solnet.Programs.Test/TokenProgramTest.cs
```diff
@@ -792,5 +919,501 @@ public void TestThawAccountMultiSignature()
...
+            Assert.IsTrue(decodedInstructions[1].Values.TryGetValue("Destination", out object destination));
+            Assert.IsTrue(decodedInstructions[1].Values.TryGetValue("Authority", out authority));
+            Assert.IsTrue(decodedInstructions[1].Values.TryGetValue("Signer 1", out signer1));
+            Assert.IsTrue(decodedInstructions[1].Values.TryGetValue("Signer 2", out signer2));
+            Assert.IsTrue(decodedInstructions[1].Values.TryGetValue("Signer 3", out signer3));
+            Assert.AreEqual("6ULjh7zKSsbiMqNzpCKMgYCUDFZ9Thxy59SGE48x22AF", (PublicKey)source);
+            Assert.AreEqual("5omQJtDUHA3gMFdHEQg1zZSvcBUVzey5WaKWYRmqF1Vj", (PublicKey)destination);
+            Assert.AreEqual("6yg3tZM1szHj752RDxQ1GxwvkzR3GyuvAcH498ew1t2T", (PublicKey)signer1);
+            Assert.AreEqual("88SzfLipgVTvi8hQwYfq21DgQFcABx6yAwgJH5shfqVZ", (PublicKey)signer2);
+            Assert.AreEqual("5Xcw7EQb6msgpVdGB8Hf8kpCqVyacTChgFBUphpuUeBo", (PublicKey)signer3);
```
tiago18c, line 1416: Question: aren't these casts to `PublicKey` redundant? IIRC the `AreEqual` method accepts `object` type.

tiago18c (changes requested):
Please fix

test/Solnet.Programs.Test/TokenProgramTest.cs
murlokito: Yeah I think it would've worked now that there's the `ToString()` previously it wouldn't work as well iirc

tiago18c (approved):
LFG;

Issue #174, opened 2021-08-06
Make TransactionInstruction construction accessible to non C# languages


liquizard:
**Is your feature request related to a problem? Please describe.**
Some .Net languages like VB.Net do not support init only setters.
This prevents the construction of TransactionInstruction instances.

**Describe the solution you'd like**
A simple workaround to allow the creation of TransactionInstruction via a Create() method.
Suggest a separate Factory class to keep this Create method away from the main way users would interact with this class.
e.g.

    TransactionInstructionFactory.Create(_program_id.KeyBytes(), keys, data)

Pull request incoming.

**Describe alternatives you've considered**
Changing properties etc. - too much impact for what is a minority edge case.

murlokito:
I'm assuming this is for instructions related to programs we haven't implemented yet?

liquizard:
Howdy - yes, I'm reverse engineering Tulip harvest and ran into this issue when creating the a program class in my prototyping.

murlokito:
Alright sounds good to me. I'm wondering if the same issue affects the programs we have already implemented but I'm assuming they work fine?

murlokito:
Also, if you're down to make it something robust and to potentially releasing it on NuGet, we'd be happy to eventually add it to the list of ecosystem libraries in the README. We'd like to have a cool list of program implementations which aren't using Anchor, since in due time we'll have code generation for Anchor based programs :)

liquizard:
The issue does not affect your built-in programs. 
They work fine, as they have their own factory methods, e.g.

        Dim builder As New TransactionBuilder()
        builder.
            SetRecentBlockHash(blockHash.Result.Value.Blockhash).
            SetFeePayer(fromAccount).
            AddInstruction(MemoProgram.NewMemo(fromAccount, "Hello from Sol.Net :)")).
            AddInstruction(SystemProgram.Transfer(fromAccount, toAccount, 100000))

This issue only crops up when implementing anything that needs to create its own TransactionInstruction instances. 

Thanks - would be happy to share.
Got harvest working, still trying to fathom the enumeration of vaults that belong to an account like Step does.
Any pointers would be greatly appreciated :-)

murlokito:
Would need to take a look at the Tulip structures to know for sure, but I'm assuming the "owner" account is referenced somewhere in the structure, in that case I think you can use `GetProgramAccounts` with the Tulip address and filtering by the offset and comparing with the owner's address at that given offset.

[...]

PR #175, opened 2021-08-06
Adds TransactionInstructionFactory

liquizard:
| Ready | Workaround | Yes (addition) | #174 

## Problem
See #174 

## Solution
Added a separate TransactionInstructionFactory class with static Create method.
Not sure whether I've picked the right namespace, don't mind if you want to relegate this to a Helpers/Utilities class/namespace
Let me know if you'd like this moving/renaming.

Thanks!

### DIFF
--- a/src/Solnet.Rpc/Builders/TransactionInstructionFactory.cs
+++ b/src/Solnet.Rpc/Builders/TransactionInstructionFactory.cs
@@ -12,8 +12,8 @@ namespace Solnet.Rpc.Builders
     {
 
         /// <summary>
-        /// Helper method to make TransactionInstruction objects available to languages
-        /// that do not currently support immutable init-only setters like VB.Net
+        /// Helper method to make TransactionInstruction objects available to outside Program implementations in 
+        /// languages that do not currently support immutable init-only setters like VB.Net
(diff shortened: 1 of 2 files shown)


src/Solnet.Rpc/Builders/TransactionInstructionFactory.cs
```diff
@@ -0,0 +1,41 @@
...
+    /// </summary>
+    public static class TransactionInstructionFactory
+    {
+
+        /// <summary>
+        /// Helper method to make TransactionInstruction objects available to outside Program implementations in 
+        /// languages that do not currently support immutable init-only setters like VB.Net
+        /// </summary>
+        /// <param name="programId"></param>
+        /// <param name="keys"></param>
```
murlokito, line 19: Would just like to see these comments here, looks good to me otherwise :)

murlokito (approved):
Just that one comment and will merge it :)

src/Solnet.Rpc/Builders/TransactionInstructionFactory.cs
liquizard: Sorted ✔️

PR #178, opened 2021-08-08
Added new flags and reworked old tests to use reflection

### LINKED ISSUE #177: Add flags which use different values for masking
In the context of `Solnet.Serum` due to the recent addition of the Permissioned Markets feature, three new bit masks were [added](https://github.com/project-serum/serum-dex/blob/master/dex/src/state.rs#L59) to the `AccountFlags` structure, making the current `ByteFlag` no longer enough to hold all of the possible values.

Proposing additions:
- `ShortFlag` as an intermediary
- `LongFlag` as a long term safe bet

murlokito:
| Read | Feature/Refactor | Yes | Closes #177  |

## Problem

See #177

## Solution

Added new types of flags using `Flag<T>` and reworked the `ByteFlag` tests to use reflection

### DIFF
--- a/src/Solnet.Programs/Abstract/ByteFlag.cs
+++ b/src/Solnet.Programs/Abstract/ByteFlag.cs
@@ -8,42 +8,42 @@ namespace Solnet.Programs.Abstract
         /// <summary>
         /// Check if the 1st bit is set.
         /// </summary>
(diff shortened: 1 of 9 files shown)


tiago18c (approved):
LGTM;

Issue #176, opened 2021-08-08
Add TokenWallet + Solnet.Extensions project


liquizard:
**Is your feature request related to a problem? Please describe.**
I'm thinking there is a gap for a higher-level `TokenWallet` construct that is also capable of understanding and navigating the SPL sub-account relationships of a particular on-chain wallet. This would use the primitive Wallet and RPC together to load account information and balances etc. Such an object could be interrogated by a developer to find Associated Token Accounts for a mint and provision an ATA or sub-account if necessary. 

The Solnet.Wallet project provides all the primitive elements required to interact with a wallet and accounts. As this is all based on crypto and maths all this functionality is deterministic.

The Solnet.Rpc project provides all the means of discovering and querying the blockchain to find out the current real world specifics of a particular wallet, its balance, owned SPL token accounts etc.

This TokenWallet behavior is not mathsy deterministic wallet logic as it depends on the RPC and on-chain state. More akin to RPC functionality as it is mockable for testing inline with other RPC tests. But then it is not primitive RPC functionality either - it just makes use of the RPC. 

For example:

    var wallet = TokenWallet.LoadSnapshot(client, publicKey); // discover accounts via getTokenAccountsByOwner
    var ata = wallet.GetAssocatedTokenAccount(tokenAddress); // get account or null if none exist
    if (ata == null) ata = wallet.CreateAssociatedTokenAccount(tokenAddress); // provision ATA
    var balance = wallet.GetConsolidatedBalance(tokenAddress); // sum balanced of all accounfs for SPL token
    ...
    wallet.Refresh();

Looking to the solanaj project for answers, there is a `TokenManager` class that makes use of the RPC that I think is the closest looking animal. It sits in the `solanaj.token` namespace. Their project doesn't seem to have internal project dependency boundaries to contend with like C#. I couldn't find anything that provides a consolidated balance of a particular SPL token for a top-level account. Is this something we would anticipate a developer should be managing themselves as a necessary complexity?

I have the fundamentals of `TokenWallet` and would be happy to contribute this back into Solnet, but where would it belong? Going off the solonaj namespace, the closest I can come up with would be `Solnet.Token` - but this would need to be a new project. or is that overkill?

murlokito:
This is a really nice add and something I'd love to have in Solnet, I've previously done something similar as a wholly different project to manage ATAs in an application. I'm just a bit fuzzy on the namespace, but at the same feels like `Solnet.Token` is probably the best bet.

[...]

PR #179, opened 2021-08-10
Solnet.Extensions and TokenWallet

liquizard:
| Ready to Merge (when you're happy) | Feature | Yes (additions) | #176 |

Howdy, here's a work-in-progress PR .
I could do with some steering advice please to avoid any going to far down any rabbit holes.

There is a side-quest involving resolving token names and symbols. I've incorporated the fetching and parsing of the Solana standard token list used by the ecosystem. Not sure whether this is valid scope for this `Solnet.Extensions` project, or should be a separate thing is in its own right but I want to avoid any additional external dependencies.

Progress so far:
- `Solnet.Extensions` project added
- `TokenWallet` created + getting token accounts and balances
- `TokenInfo` resolving via Solana token list fetch/parse
- Examples updated to show usage
- Associated Token Account indicator

You can take `TokenWallet` for a spin with the example program.

Some questions:
- Is the organization and naming of `Solnet.Extensions` agreeable?
- `TokenInfo` and Solana Token List - is this inline with the wider goals of Solnet?
- Is it friendlier for developers to interact with addresses as strings or use `PublicKey` everywhere? (gone with string for now)
- What approach would you like taken for unit testing `TokenWallet`? Lots of hard work has already gone in for coverage. I was planning to inject a mock RPC client with some mock responses to exercise the codebase - would that be ok?
- What exception strategy would you prefer for `TokenWallet.Refresh()` - if RPC methods are not successful - are you happy for me to throw a specific `TokenWalletException` ?


murlokito:
I really like where this is going, me and @tiago18c also discussed adding structures for the Solana token list and I'm very happy to see you also went for it along with the `TokenWallet`, feels very much in line with what we want for Solnet :)

Just a few thoughts I had from looking at the PR so far:
- I saw you hardcoded the `Tokenkeg....` public key again, but you also have `Solnet.Programs` as dependency for ATA, you can just use `TokenProgram.ProgramIdKey`, no need to be hardcoding the key again.
- If `Solnet.Extensions` is really going to grow with more additions in the future I'd like to keep seeing a certain degree of separation between models and the classes which provide most of the functionality, i.e. having `ITokenInfoResolver`, `TokenMintResolver` and `TokenWallet` at the root of the project while the rest of the classes are moved into `Models/TokenWallet/`, `Models/TokenInfo/`
- There are tokens in the token list which have duplicate (even more than 2 instances, at times) of the same mint, this might be something to take into account.

Those are the things that came to mind from the WIP.

To answer your questions:
- Naming I can agree with it as it stands, just perhaps having `Solana` in the name of the mint list/token feels weird, in a private project where I also did this I just called them `TokenList` and `TokenInfo`, but I see you're doing a certain separation and it's ok, in terms of organization what I said above answers it, I think
- Regarding `PublicKey` vs `string`, for something like this I don't think going with string is a problem, when using the `TokenWallet` I'm assuming developer will end up using both, and we have implemented operators for the keys exactly to make this easier.
- Unit testing `TokenWallet`, yes, injecting a mock is what we've been doing to test everything that requires an RPC request in the workflow
- We've been attempting to keep exception-based logic to a minimum but I think I see where you're going with that and I think it's fine.

One question from me is, in the context of what you're doing where you'd like these additions, would it make a difference to use Asyncs? This is just because these additions really remove a lot of "boilerplate" code from potential desktop wallet implementations and UIs would greatly appreciate the use of the Async methods.

Some of these comments are just :nail_care: and don't matter too much, overall I like the ideas and am looking forward to :ship: this

liquizard:
Thanks for the feedback.
Yes to Asyncs - wanted to get the basics outlined before following up with asyncs and comments etc.

murlokito:
Excellent! Keep up the great work :)

liquizard:
Hello there. More work in progress.
From a callee perspective, SPL send is now down to a handful of lines with much of the ATA complexity abstracted away.
e.g.

        var logger As New MyAppLogging();
        var client = ClientFactory.GetClient(Cluster.MainNet, logger);
        var tokens = TokenInfoResolver.Load();

        // find source of funds
        var wallet = TokenWallet.Load(client, tokens, owner);
        var sourceAccount = wallet.TokenAccounts.WithSymbol("SRM").WithAtLeast(5.5D).FirstOrDefault();

        // send 5.5 SRM to target wallet
        var sig = wallet.Send(sourceAccount, 5.5D, targetWalletPublicKey, owner);
        Console.WriteLine(sig.Result);

Wondering whether to bury the FirstOrDefault inside the `wallet.Send` method for neatness, but think it might be misleading to a callee for the Send method to accept multiple sourceAccounts to only use the first one.

The callee can also reuse the ATA account logic passing their own `TransactionBuilder` into `JitCreateAssociatedTokenAccount` if needed.

Couple more questions if you don't mind:
- Are you ok with `TokenWallet.Send` methods returning RPC client `RequestResult<string>` responses directly?
- Are there any existing utility methods I should be using to convert decimal amounts into raw ulongs?
- What are your thoughts on me adding a WellKnownTokens reference class with a dozen or so mainstream tokens? I think this could save a lot of unnecessary traffic to the token list and make it easier code samples much cleaner.
- Thanks for the heads-up on token duplicates in the token list - any preference on which should take priority? Keep first and ignore the rest?

murlokito:
This is looking really, really good. I think you shouldn't bury the `FirstOrDefault()` inside the `Send`, it should only receive the `sourcePublicKey`/`sourceAccount`

Regarding the questions:
- Yes, perfectly ok with that, it's neat so you can `TokenWallet.Send` and then fire up a `SignatureSubscribe` in the background with the returned signature, or whatever the callee desires to do.
- There are no utility methods regarding decimal amounts <-> raw ulongs yet, I'm assuming you're asking because of converting from the friendly value into the value you should pass into the instruction. I've done something in the context of Serum because of order placement but I'm fairly sure it's very different because it involves lot sizes.
- Perfectly fine with that.
- This really depends, the list contains everything from NFTs to LPTs, you can filter those out by the tags and possibly by some info in the extensions. I think, for the SPL token cases, we should prefer those with `chainId = 101`, in the cases where I saw that the token's mint was duplicated it was the occurrence with that chainId that had most of the info, i.e, address of serum markets, coingeckoId and other links.

liquizard:
Thanks for the answers.

You're right - the conversion is for dealing with user/dev friendly decimals to raw ulongs. If there's nothing around for that, I'll extract my solution into a utility class/method. Any preference on naming and where that belongs?

I think a Solnet curated list of well known tokens, with the option for the callee to inject their own tokens or load the full token list is a good way to go and creates an opportunity to shout about new tokens making the curated list :-)

murlokito:
Yes, I can agree with that. Would love to have a static class to help with the conversions but honestly I'm not really sure where it should be added, if Solnet.Programs or on Extensions as well.. I think either way it's something important to have so I'd just go ahead and add it somewhere and we can see later if it makes sense or not

liquizard:
Added `WellKnownTokens` with a handful of familiars for now.
On naming the tokens and readability, which do you prefer?

                                A                      or                               B ?
    wallet.TokenAccounts.ForToken(WellKnownTokens.SRM) vs wallet.TokenAccounts.ForToken(WellKnownTokens.Serum)
    wallet.TokenAccounts.ForToken(WellKnownTokens.MNGO) vs wallet.TokenAccounts.ForToken(WellKnownTokens.Mango)
    wallet.TokenAccounts.ForToken(WellKnownTokens.RAY) vs wallet.TokenAccounts.ForToken(WellKnownTokens.Raydium)
    wallet.TokenAccounts.ForToken(WellKnownTokens.TULIP) vs wallet.TokenAccounts.ForToken(WellKnownTokens.Solfarm)
 
This lets you to filter the token accounts with a mint using readable code like this...

    var wallet = TokenWallet.Load(client, tokens, owner);
    var rayAta = wallet.TokenAccounts.ForToken(WellKnownTokens.RAY).
                                           WhichAreAssociatedTokenAccounts().FirstOrDefault();

Open to suggestions on the naming of the `WhichAreAssociatedTokenAccounts` method.
This does what it says on the tin, but wondered if you can suggest a snappier name.

I slept on the whole `FirstOrDefault` conundrum, and added this simplification which resolves the filter list down to a single ATA or null. This should be used in conjunction with a `ForToken`, `WithMint` or `WithSymbol` selector otherwise you would get a random ATA from your wallet which would not be useful. Code sample above becomes:

    var wallet = TokenWallet.Load(client, tokens, owner);
    var rayAta = wallet.TokenAccounts.ForToken(WellKnownTokens.RAY).AssociatedTokenAccount();

The need to download the token list is gone also unless wanted and `TokenInfoResolver` can be bootstrapped by a developer with their own tokens as you can see in the example. 

    var tokens = new TokenInfoResolver();
    tokens.Add(new TokenDef("AHRNasvVB8UDkU9knqPcn4aVfRbnbVC9HJgSTBwbx8re", "Solnet Test Token", "STT", 2));

murlokito:
That's awesome! Don't really have a preference about naming there, I reckon you can keep it with symbols if that's what you currently have. `WellKnownTokens.SRM`

liquizard:
Opted for B, the longer form as I think it looks for more readable code, e.g.

    var account = wallet.TokenAccounts.ForToken(WellKnownTokens.Raydium).AssociatedTokenAccount();

liquizard:
Recent updates
- Tests added for `Solnet.Extensions.Test` for `TokenResolver` and `TokenWallet`
- Ditched Moq and built queued response mock RPC in `MockTokenWalletRpc` - as discussed this may offer a mocking strategy you could apply in `Solnet.Serum`
- To make mockery manageable introduced `ITokenWalletRpcProxy` and `TokenWalletRpcProxy` that houses RPC methods used by `TokenWallet` allowing an injectable mock to be created in `Solnet.Extensions.Test`
- Some more overloads, minor renames and tidy-ups for consistency
- Introduced a couple of additional list filters and validations
- Merged latest main branch in too so shouldn't be too risky to take this branch for a spin now

Sending tokens now looks like this:

        var client = ClientFactory.GetClient(Cluster.MainNet, logger);
        var tokens = New TokenInfoResolver();
        var wallet = TokenWallet.Load(client, tokens, feePayer);

        // find source of funds
        var source = wallet.TokenAccounts.ForToken(WellKnownTokens.Serum).WithAtLeast(11.75D).FirstOrDefault();

        // execute
        var sig = wallet.Send(source, 12.75D, target, feePayer);

Current Limitations
- Cannot send native SOL via the `Send` method.
- `TokenWallet.Load` assumes a SOL-native top-level account is provided.
To cosset the callee, I think the error detection needs to be bullet proof given the scope of address kinds that could be provided.
The `TokenWallet.Load` method needs to be fuzzed with different types of public key - program, PDA, existing ATA etc. and barf under certain circumstances and cope in others. It may be valid for Send in some cases such as direct to ATA for example I think should be allowed for matching mints.

murlokito:
This is looking very, very good, will take it for a spin later today.

Regarding the native SOL Send limitation, I don't think that's an issue, the goal of this is to be a `TokenWallet` and it seems to be fulfilling that role quite well.

Regarding the Load method being fuzzed I think I may have an idea, in `Solnet.Rpc.Utilities` there is an `Ed25519Extensions` class which implements an extension method called `IsOnCurve`, this method is currently internal, it should be made public, along with it's class. the idea for this is that `TokenWallet.Load` in a case such as the following (copy pasta'd from your):

```csharp 
var wallet = TokenWallet.Load(client, tokens, feePayer);
```

`feePayer` should be converted to `PublicKey` and `IsOnCurve` should be called on it's `byte[]`, this makes us know for a fact that the passed `feePayer` isn't either a PDA or an ATA, because ATA and PDA generation requires the generated addresses to not be on the curve.

liquizard:
On The Curve sounds like a great solution - wasn't aware of that, thanks for the tip.
That's gonna sniff out my made-up test keys and wreck the test - but c'est la vie.

Re: docs + comments - I understand, all part of the process. I'll be sorting those once we're happy with the additions.

murlokito:
Also re docs, and related to #185, since you're working on this it would be awesome if the docs could have an addition of yours to give more context on how to use these additions, it would be legendary, really.

liquizard:
Howdy, I think we're just about there and ready to merge.
Comments added and the `IsOnCurve` validation incorporated into `TokenWallet.Load` and `Send` methods.

liquizard:
If it's not going to cause any conflicts with on going work I'll add some info to the repo's README.

murlokito (approved):
Took this for a spin yesterday evening and it was looking mighty fine, props for this! Will wait for @tiago18c to check and approve as well :)

liquizard:
Good morning,

I've introduced a primitive `TokenQuantity` object that performs the primitive decimal/raw conversions and allows a mint, symbol, decimal places, name and quantities to be passed around the place. Used like this:

```
            var qty = WellKnownTokens.USDC.CreateQuantity(4741784U);
            Assert.AreEqual(4741784U, qty.BalanceRaw);
            Assert.AreEqual(4.741784M, qty.BalanceDecimal);
            Assert.AreEqual("USDC", qty.Symbol);
            Assert.AreEqual(6, qty.DecimalPlaces);
            Assert.AreEqual("4.741784 USDC (USD Coin)", qty.ToString());
```

I'm thinking a couple of minor renames are in order, but want to coordinate these with you as they may break anything that you may have tried this branch out.

I'm proposing the following:
- `ITokenInfoResolver` --> `ITokenMintResolver`
- `TokenInfoResolver` --> `TokenMintResolver`
- Two property name changes on the `TokenQuantity` class hierarchy, from `BalanceRaw` to `QuantityRaw` and `BalanceDecimal` to `QuantityDecimal`

These are better descriptions for what's going on now.

murlokito:
Really like those additions of the Token Quantity, feel free to rename the properties and classes, it will break things on my side for the app but it's not a big deal, this is why we should dog feed our own APIs to sample apps and similar things to make sure things are following the proper convention etc.

liquizard:
Thinking about the issue of passing Account into library objects raised #192 and the goal of producing trustable library code the `TokenWallet.Send` method should be reworked to follow best practice in this area - specifically implement `RequestSignature` callback and follow the pattern established in the Serum library. This way the object's internals would not require blind trust of any kind.

liquizard:
Removed all traces of `Account` requirements from TokenWallet so it never needs to know your private keys.
All variants of Send require a `Func<TransactionBuilder>, byte[]> signTxCallback` function passed in for the caller to sign the transaction prior to send.

The existing `JitCreateAssociatedTokenAccount` method can be used to add the associated token account provisioning instructions to an existing `TransactionBuilder` instruction.

This is how calling token send code now looks:

```csharp
var feePayer = signer.PublicKey;
var tokenAccount = wallet.TokenAccounts().ForToken(WellKnownTokens.Raydium).WithAtLeast(5M).First();
wallet.Send(tokenAccount, 1M, targetOwner, feePayer, builder => builder.Build(signer));
```
This fix is part of the Creating Trustable Library Code idea raised in #192.

murlokito:
Love to see these latest changes. @tiago18c can you please review this so we can merge these two open PRs and release v0.4.4?

liquizard:
Paging @tiago18c... 
Please could you review and consider merging for release as 0.4.4?

src/Solnet.Extensions/Models/TokenWallet/TokenWalletFilterList.cs
```diff
@@ -0,0 +1,152 @@
...
+        /// <summary>
+        /// Keeps all accounts with at least the supplied minimum balance.
+        /// </summary>
+        /// <param name="minimumBalance">A minimum balance value as decimal.</param>
+        /// <returns>A filtered list of accounts with at least the balance as decimal supplied.</returns>
+        public TokenWalletFilterList WithAtLeast(decimal minimumBalance)
+        {
+            return new TokenWalletFilterList(_list.Where(x => x.QuantityDecimal >= minimumBalance));
+        }
+
```
tiago18c, line 105: I would add a variant of this filter for > 0

`public TokenWalletFilterList WithNonZero()`

Would be very useful in a wallet UI to display tokens with any value :)

src/Solnet.Extensions/WellKnownTokens.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// <summary>
+        /// SRM (Serum)
+        /// </summary>
+        public static TokenDef Serum = new TokenDef("SRMuApVNdxXokk5GT7XD5cUUgXMBCoAz2LHeuAoKWRt", "Serum", "SRM", 6);
+
+        /// <summary>
+        /// RAY (Radium)
+        /// </summary>
+        public static TokenDef Raydium = new TokenDef("4k3Dyjzvzp8eMZWUXbBCjEvwSkkk59S5iCNLY3QrkX6R", "Raydium", "RAY", 6);
+
```
tiago18c, line 40: should we add a few more? MNGO, BTC, ETH

tiago18c (approved):
Looks very good.
Just not a great fan of the proxy thing. In my view, it adds another layer of abstraction, and the tests could be achieved by mocking the existing IRpcClient (@murlokito got it working properly for multiple sequential requests on the Serum library). But it seems pointless to rework that now.

I say ship it as is :)

src/Solnet.Extensions/WellKnownTokens.cs
liquizard: Yep - got a load ready to go...
- Liquid Protocol
- Boring Protocol
- Fabric
- Samoyed
- Tulip
- Saber
- Mango
- Orca
- Only1
- Step 

Any others you'd like to see?

liquizard:
Thanks for the feedback!

> Just not a great fan of the proxy thing...
Yeah - it didn't sit right with hoak either which I take on board. I've got some RPC pre-request/post-response hooks in the works that will come in very useful for cross-concerns like rate limiting etc. With these hook points in place, it will be straightforward to refactor the proxy away with a hooked-RPC. Happy to chalk this up for simplification once that's in place to migrate to.

Let me know if there's any Well Known Tokens you'd like adding, the we can ship this.

murlokito:
This has already been shipped, we can just add improvements going on

clintnetwork:
> ```cs
> builder => builder.Build(signer)
> ```

This should be better documented!

murlokito:
> > ```cs
> > builder => builder.Build(signer)
> > ```
> 
> This should be better documented!

I don't get it. There's plenty of examples/documentation on the TransactionBuilder, plus it has nothing to do with this PR so why this comment?

clintnetwork:
> > > ```cs
> > > builder => builder.Build(signer)
> > > ```
> > 
> > 
> > This should be better documented!
> 
> I don't get it. There's plenty of examples/documentation on the TransactionBuilder, plus it has nothing to do with this PR so why this comment?

Sorry github quote replied not so good, I was speaking about the `.Send()` method that he's described in different examples on the repo without the transaction callback so I was lost about what to do

PR #181, opened 2021-08-11
Fix get program accounts edge case

### LINKED ISSUE #180: Fix edge case in GetProgramAccounts when dataSize isn't passed
Because of the way we're filtering the parameters so as to not send anything when there are no extra parameters, when the `dataSize` isn't passed, the method blows up because `filters` remains null.

hoakbuilds:
| Ready | Hotfix | No | Closes #180  |

## Problem

See #180 

## Solution

Added a nullcheck for the `filters` whenever the `memCmpList` isn't null.


tiago18c (approved):
LGTM;

PR #197, opened 2021-08-28
Adding Memo Program V2 address and flag to check if it's v1 or v2

### LINKED ISSUE #196: Add Memo Program V2 address
### Problem
The [Memo Program](https://github.com/solana-labs/solana-program-library/blob/master/memo/program/src/lib.rs#L21) has for a while now been upgraded to V2, we only have the current V1 address and thus V2 instructions aren't being correctly tagged and decoded.

### Solution
We should add the V2 address.

murlokito:
| Ready | Feature | No | Closes #196  |

## Problem

See #196 

## Solution

Added the address for Memo Program V2 and a flag to check which address
Added `NewMemoV2` so as not to break the many tests and existing implementations of `NewMemo`, made it so that it's not required to pass a `PublicKey` for the message to be signed.

Tested [here](https://explorer.solana.com/tx/2uRBzW3EqJRA2ADWbKQPCp1yfKvQ8Bh74jqeuc6CRYyakF9xWxejMAeHBuGCWwF1rK4KVwERe6cq9R2nVPCvTWFG?cluster=testnet)

### DIFF
--- a/src/Solnet.Examples/HelloWorldExample.cs
+++ b/src/Solnet.Examples/HelloWorldExample.cs
@@ -40,7 +40,7 @@ namespace Solnet.Examples
 
                     Console.WriteLine($"Balance: {balance.Result.Value}");
 
-                    var memoInstruction = MemoProgram.NewMemo(wallet.Account, "Hello Solana World, using Solnet :)");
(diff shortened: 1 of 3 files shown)


tiago18c (approved):
LGTM;

Issue #213, opened 2021-09-21
Batching RPC requests


murlokito:
The RPC allows batching of RPC requests, they should be submitted as an array of the serialized RPC requests.

PR #221, opened 2021-10-04
JsonRpcBatch request/response and composer

liquizard:
| In Progress | Feature | Yes (enhancement) | #213 |

## Background
"Requests can be sent in batches by sending an array of JSON-RPC request objects as the data for a single POST."
From: https://docs.solana.com/developing/clients/jsonrpc-api

## Solution
This is a working implementation that moves us towards solving #213 
Created `JsonRpcBatchRequest`, `JsonRpcBatchResponse` and `JsonRpcBatchResponseItem` for serialization and deserialization of batches of requests and responses in an abstract way.
Created `SolanaRpcBatchComposer` to create the batch of requests and coordinate the concrete data type of the requests and responses.

`TokenWallet` is a good first candidate for migration onto batched requests as we can combine the `GetBalance` and `GetTokenAccountsByOwner` requests into a single request.

The performance improvement on large request batches looks promising.

Example usage:

```csharp
// compose a new batch of requests
var batch = new SolanaRpcBatchComposer();
batch.GetBalance("9we6kjtbcZ2vy3GSLLsZTEhbAqXPTRvEyoxa8wxSqKp5");
batch.GetTokenAccountsByOwner("9we6kjtbcZ2vy3GSLLsZTEhbAqXPTRvEyoxa8wxSqKp5", null, TokenProgram.ProgramIdKey);
batch.GetConfirmedSignaturesForAddress2("9we6kjtbcZ2vy3GSLLsZTEhbAqXPTRvEyoxa8wxSqKp5", 200, null, null);
batch.GetConfirmedSignaturesForAddress2("88ocFjrLgHEMQRMwozC7NnDBQUsq2UoQaqREFZoDEex", 200, null, null);
batch.GetConfirmedSignaturesForAddress2("4NSREK36nAr32vooa3L9z8tu6JWj5rY3k4KnsqTgynvm", 200, null, null);

// execute batch
var response = batch.Execute(client);
```

## Next Up
The above sample shows the handling of the response in batch after the execute method. 

There are opportunities to design more elogant ways to compose a batch combining callbacks or using the async/task pattern for each method, e.g.

```csharp
// supply a callback 
batch.GetConfirmedSignaturesForAddress2("9we6kjtbcZ2vy3GSLLsZTEhbAqXPTRvEyoxa8wxSqKp5", 200, null, null, 
      x => {
          // use x
      });

// async/task model
var result = batch.GetConfirmedSignaturesForAddress2Async("9we6kjtbcZ2vy3GSLLsZTEhbAqXPTRvEyoxa8wxSqKp5", 200, null, null);
```

The crudely named `RpcBatchReqRespItem` (working title) which handles the association of each request and its return type could be extended to handle callbacks or tasks.

IMO there are at least a couple of different directions we could go in to best cater for different skill levels of the developers interacting with the library. It might be better to offer more complex callback/async wrappers on top of a simple batch handling kernal such as this composer. 

As of now, the `SolanaRpcBatchComposer` supports a small PoC subset of the full RPC client methods, once the approach is proven, the remaining methods can be added.


liquizard:
Added callbacks for batch methods. e.g.

```csharp
// compose a new batch of requests
var batch = new SolanaRpcBatchComposer();
batch.GetBalance("9we6kjtbcZ2vy3GSLLsZTEhbAqXPTRvEyoxa8wxSqKp5",
    callback: x => found_lamports = x.Value);
batch.GetTokenAccountsByOwner("9we6kjtbcZ2vy3GSLLsZTEhbAqXPTRvEyoxa8wxSqKp5", null, TokenProgram.ProgramIdKey,
    callback: x => found_balance = x.Value[0].Account.Data.Parsed.Info.TokenAmount.AmountDecimal);
batch.GetConfirmedSignaturesForAddress2("9we6kjtbcZ2vy3GSLLsZTEhbAqXPTRvEyoxa8wxSqKp5", 200, 
    callback: x => sig_count += x.Count);
batch.GetConfirmedSignaturesForAddress2("88ocFjrLgHEMQRMwozC7NnDBQUsq2UoQaqREFZoDEex", 200, 
    callback: x => sig_count += x.Count);
batch.GetConfirmedSignaturesForAddress2("4NSREK36nAr32vooa3L9z8tu6JWj5rY3k4KnsqTgynvm", 200, 
    callback: x => sig_count += x.Count);
```

There are many different execution and concurrency options to be considered as well as callback semantics using tasks/asyncs in creating a general multi-purpose library. This callback arrangement was the simplest solution to achieve my goals working with the batch composer.

For simplicity and thread-safety, callbacks are invoked sequentially by `SolanaRpcBatchComposer.ProcessBatchResponse` method. This is decoupled and invocable separately from the `Execute` method for testability.

liquizard:
This latest update is a long way off the Batch RPC's final form, but presents two competing ideas and shows the different calling semantics of Asyncs vs. Callbacks. The commit contains two equivalent batch API's, `SolanaRpcBatchWithAsyncs` and `SolanaRpcBatchWithCallbacks` that each wrap the underlying `SolanaRpcBatchComposer` object.

Compare two implementations of the same logic:

## Async 
The `SolanaRpcBatchWithAsyncs` object is intended to be a drop-in replacement for the `SolanaRpcClient` supporting the SomeMethodAsync method convention. For the batch use case, tasks brings with them a big hazard. Any call to a `task.Result` before the batch execution will cause immediate deadlock. So the opportunity to realistically swap this in for the majority of existing code that uses `SolanaRpcClient` is likely to be unsuccessful. Safely using `SolanaRpcBatchWithAsyncs` leads to code where all tasks must be kept and then processed after the call to `batch.Execute`. 

```csharp
// compose a new batch of requests
var batch = new SolanaRpcBatchWithAsyncs();
var balance = batch.GetBalanceAsync("9we6kjtbcZ2vy3GSLLsZTEhbAqXPTRvEyoxa8wxSqKp5");
var tokensAccounts = batch.GetTokenAccountsByOwnerAsync("9we6kjtbcZ2vy3GSLLsZTEhbAqXPTRvEyoxa8wxSqKp5", null, TokenProgram.ProgramIdKey);
var sigResults1 = batch.GetConfirmedSignaturesForAddress2Async("9we6kjtbcZ2vy3GSLLsZTEhbAqXPTRvEyoxa8wxSqKp5", 200, null, null);
var sigResults2 = batch.GetConfirmedSignaturesForAddress2Async("88ocFjrLgHEMQRMwozC7NnDBQUsq2UoQaqREFZoDEex", 200, null, null);
var sigResults3 = batch.GetConfirmedSignaturesForAddress2Async("4NSREK36nAr32vooa3L9z8tu6JWj5rY3k4KnsqTgynvm", 200, null, null);

// process and invoke callbacks
batch.Composer.Execute(client);

// process task results
found_lamports = balance.Result.Value;
found_balance = tokensAccounts.Result.Value[0].Account.Data.Parsed.Info.TokenAmount.AmountDecimal;
sig_callback_count += sigResults1.Result.Count;
sig_callback_count += sigResults2.Result.Count;
sig_callback_count += sigResults3.Result.Count;
```

## Callbacks
The `SolanaRpcBatchWithCallbacks` object borrows from the node.js pattern of providing callbacks to methods. The benefit is that code for processing the response is near to the method invocation and we can avoid keeping hold of the tasks for later use.

```csharp
// compose a new batch of requests
var batch = new SolanaRpcBatchWithCallbacks();
batch.GetBalance("9we6kjtbcZ2vy3GSLLsZTEhbAqXPTRvEyoxa8wxSqKp5",
    callback: (x, ex) => found_lamports = x.Value);
batch.GetTokenAccountsByOwner("9we6kjtbcZ2vy3GSLLsZTEhbAqXPTRvEyoxa8wxSqKp5", null, TokenProgram.ProgramIdKey,
    callback: (x, ex) => found_balance = x.Value[0].Account.Data.Parsed.Info.TokenAmount.AmountDecimal);
batch.GetConfirmedSignaturesForAddress2("9we6kjtbcZ2vy3GSLLsZTEhbAqXPTRvEyoxa8wxSqKp5", 200, null, null,
    callback: (x, ex) => sig_callback_count += x.Count);
batch.GetConfirmedSignaturesForAddress2("88ocFjrLgHEMQRMwozC7NnDBQUsq2UoQaqREFZoDEex", 200, null, null,
    callback: (x, ex) => sig_callback_count += x.Count);
batch.GetConfirmedSignaturesForAddress2("4NSREK36nAr32vooa3L9z8tu6JWj5rY3k4KnsqTgynvm", 200, null, null,
    callback: (x, ex) => sig_callback_count += x.Count);

// process and invoke callbacks
batch.Composer.Execute(client);
```

Under the category of "can open, worms everywhere" another design consideration is how best to handle exceptions. Should a single transport exception for the batch be applied to each callback?

tiago18c:
I have to say I was surprised at how quickly you came up with a solution to this. I thought we'd have to change the client and probably end up with breaking changes.

Regarding the two possible solutions, I think the callback one is cleaner and less error prone. There is no chance for the user to invoke the `Result` before the `Execute` call. As for exceptions, it's a matter of intercepting them and adding the relevant info to the relevant `RequestResult`.

Again, thank you and good job :)

liquizard:
Looking to bring this work into the main branch soon.
Been working on usability of this and just about to commit some auto-execute functionality to simply the processing of workloads.

Found that when using the code in workloops there is a common usage pattern emerging to need to check batch sizes and execute when batch size is full and then to check for partial batches after the workloop.

Came up with this tidier call pattern instead:
```csharp
        // setup batch to auto-execute batches of 200 requests
        var batch = new SolanaRpcBatchWithCallbacks();
        batch.AutoExecute(BatchAutoExecuteMode.ExecuteWithFatalFailure, client, 200); 
            
        // fetch all transactions
        var txs = new List<TransactionMetaSlotInfo>();
        foreach (var sig in sigs)
            batch.GetTransaction(sig, callback: (tx, ex) => txs.Add(tx));
        batch.Flush();
```

avuorine:
When using Blazor wasm while executing a batch: Cannot wait on monitors on this runtime. Changing Execute and ExecuteWithFatalFailure to use await rather than .Result for SendBatchRequestAsync fixes it.

tiago18c:
What's the status of this PR? Is it ready for review?

liquizard:
Howdy - totally missed that comment, sorry about that.
Not heard of Blazor wasm before...
That's good feedback from @avuorine, I'll see how I can rework this to fit with await rather than `.Result` and reach out and ask them to check - that's another dev I know of using the batch stuff 🙂
I'm pretty happy with how it works for what I've been using it for, also pleased with the simplied calling syntax since the introduction of auto-exexcute.

liquizard:
Paging @avuorine ...

Please could you take a look at the following update and report back on whether this is works for your Blazor wasm use case?

I've added `ExecuteAsync` and `ExecuteWithFatalFailureAsync` await friendly versions of the Execute methods on the `SolanaRpcBatchComposer` that avoid using .Result - hopefully you should be able to switch to these and you're good to go. The previous `.Result` syncronous methods remain but reuse the new async methods.

Would also welcome your views on the different styles of using batches - did you prefer `SolanaRpcBatchWithAsyncs` or `SolanaRpcBatchWithCallbacks` style? Have you tried the auto-execute mode?

avuorine:
> Paging @avuorine ...
> 
> Please could you take a look at the following update and report back on whether this is works for your Blazor wasm use case?
> 
> I've added `ExecuteAsync` and `ExecuteWithFatalFailureAsync` await friendly versions of the Execute methods on the `SolanaRpcBatchComposer` that avoid using .Result - hopefully you should be able to switch to these and you're good to go. The previous `.Result` syncronous methods remain but reuse the new async methods.
> 
I tested the Async methods and they work on Blazor wasm.

> Would also welcome your views on the different styles of using batches - did you prefer `SolanaRpcBatchWithAsyncs` or `SolanaRpcBatchWithCallbacks` style? Have you tried the auto-execute mode?
>
I tried both but ended up going with SolanaRpcBatchWithCallbacks. I've been using AutoExecute so would be helpful to have an Async version for that as well. 

Thank you!

liquizard:
Thanks for the reply.
I've found I prefer the callback pattern over the asyncs too.
Thinking of holding that one back in the finished branch but did not want to leave you abandoned.

There is a deliberate design simplification to make the batch flush execution block the calling thread due to the internal storage of the batch composer and reusing that for the next batch. In typing this through with here, I think there would be performance benefits to be had in taking a non-blocking approach and overlapping the calling code (adding to the batch) and the execution of batches (flushing the queued requests, executing and invoking the callbacks). My concern is could be quite possible for a caller loop to generate many batches very quickly that are all executing concurrently (e.g. batches of many slow getProgramAccounts) 

I'm wondering whether it could be solved with another auto-execute option, something like a `concurrent-batch-limit` that would guard against a runaway scenerario like describe above by blocking when the limit is reached.

Does the current blocking AutoExecute behavior function ok in your Blazor project?
Please could you share a little more about your use case so I could think through a good solution? 
Open to suggestions of course :-)

Also - is there any chance that you share a trivial `hello world` Blazor web-asm project that includes some Solnet activity?
I've never used this before and it would be great to have a test framework we could use to evaluate whether the batch code works and will help us to ship things that work for your use case first time :-)

avuorine:
> Thanks for the reply. I've found I prefer the callback pattern over the asyncs too. Thinking of holding that one back in the finished branch but did not want to leave you abandoned.
> 
> There is a deliberate design simplification to make the batch flush execution block the calling thread due to the internal storage of the batch composer and reusing that for the next batch. In typing this through with you here, I think there would be performance benefits to be had in taking a non-blocking approach and overlapping the calling code (adding to the batch) and the execution of batches (flushing the queued requests, executing and invoking the callbacks). My concern is quite possible for a caller loop to runaway generating many batches faster than they execute concurrently (e.g. batches of many slow getProgramAccounts)
> 
> I'm wondering whether it could be solved with another auto-execute parameter, something like a `concurrent-batch-limit` that would guard against a runaway scenerario like above by blocking when the limit is reached.
> 
> Does the current blocking AutoExecute behavior function ok as it is now in your Blazor project? Please could you share a little more about your use case so I could think through a good solution? Open to suggestions of course :-)
> 
When using AutoExecute, Flush calls the non-async versions of Execute so it fails with the same error that I posted previously.  I changed it to use the async methods instead, and it seemed to work alright on the wasm runtime after some quick testing. 

> Also - is there any chance that you share a trivial `hello world` Blazor web-asm project that includes some Solnet activity? I've never used this before and it would be great to have a test framework we could use to evaluate whether the batch code works and will help us to ship things that work for your use case first time :-)
>
I'll try to put something together in a few days so it can be tested further. 

Thanks.

tiago18c:
@liquizard been looking at this at it looks fine. Just gonna nitpick about some minor stuff. Beyond this, are we leaving both mechanisms? It seems a bit of overkill for future code maintenance, so I think it would be smart to keep just one of them and nix the other.

src/Solnet.Rpc/Core/Http/JsonRpcClient.cs
```diff
@@ -157,5 +157,106 @@ private async Task<RequestResult<T>> HandleResult<T>(JsonRpcRequest req, HttpRes
...
+                    result.RawRpcRequest = requestsJson;
+                    return result;
+                }
+
+            }
+            catch (HttpRequestException e)
+            {
+                var result = new RequestResult<JsonRpcBatchResponse>(e.StatusCode ?? System.Net.HttpStatusCode.BadRequest, e.Message);
+                result.RawRpcRequest = requestsJson;
+                // TODO - logging
```
tiago18c, line 200: Same as base Solnet
```csharp
_logger?.LogDebug(new EventId(req.Id, req.Method), $"Caught exception: {e.Message}");
```

src/Solnet.Rpc/Core/Http/JsonRpcClient.cs
```diff
@@ -157,5 +157,106 @@ private async Task<RequestResult<T>> HandleResult<T>(JsonRpcRequest req, HttpRes
...
+                return result;
+            }
+            catch (Exception e)
+            {
+                var result = new RequestResult<JsonRpcBatchResponse>(System.Net.HttpStatusCode.BadRequest, e.Message);
+                result.RawRpcRequest = requestsJson;
+                // TODO - logging
```
tiago18c, line 207: Same as base Solnet
```chsarp
_logger?.LogDebug(new EventId(req.Id, req.Method), $"Caught exception: {e.Message}");
```

src/Solnet.Rpc/Core/Http/JsonRpcClient.cs
```diff
@@ -157,5 +157,106 @@ private async Task<RequestResult<T>> HandleResult<T>(JsonRpcRequest req, HttpRes
...
+        /// <param name="response">The batch of responses obtained from the HTTP request.</param>
+        /// <returns>A task that represents the asynchronous operation that holds the request result.</returns>
+        private async Task<RequestResult<JsonRpcBatchResponse>> HandleBatchResult(JsonRpcBatchRequest reqs, HttpResponseMessage response)
+        {
+            RequestResult<JsonRpcBatchResponse> result = new RequestResult<JsonRpcBatchResponse>(response);
+            try
+            {
+                result.RawRpcResponse = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+
+                // TODO - logging - _logger?.LogInformation(new EventId(reqs.Id, reqs.Method), $"Result: {result.RawRpcResponse}");
```
tiago18c, line 228: Finish implementing the logging here and at the bottom.

src/Solnet.Rpc/SolanaRpcBatchComposer.cs
```diff
@@ -0,0 +1,384 @@
...
+
+        /// <summary>
+        /// Batch of requests and their handlers
+        /// </summary>
+        private List<RpcBatchReqRespItem> _reqs;
+
+        /// <summary>
+        /// Message Id generator.
+        /// </summary>
+        private readonly IdGenerator _idGenerator = new IdGenerator();
```
tiago18c, line 31: Maybe should expose the IRpcClient id generator as internal, and re-use it?

src/Solnet.Rpc/SolanaRpcBatchComposer.cs
```diff
@@ -0,0 +1,384 @@
...
+        }
+
+        /// <summary>
+        /// Handles the conversion of the generic JSON deserialized response objects to the native types.
+        /// </summary>
+        /// <param name="response"></param>
+        /// <returns></returns>
+        public JsonRpcBatchResponse ProcessBatchResponse(RequestResult<JsonRpcBatchResponse> response)
+        {
+            // TODO - make sure result was successful
```
tiago18c, line 160: Do we need further validation? If not delete comment.

src/Solnet.Rpc/SolanaRpcBatchComposer.cs
```diff
@@ -0,0 +1,384 @@
...
+            _reqs.Add(task);
+
+            // does this trigger an auto execute?
+            if (_autoMode != BatchAutoExecuteMode.Manual && _reqs.Count >= _autoBatchSize)
+                Flush();
+        }
+
+        private static Action<JsonRpcBatchResponseItem, Exception> WrapCallback<T>(Action<T, Exception> callback)
+        {
+            if (callback == null) return null;
```
tiago18c, line 306: Question, wouldn't it be more clean to have a callback for data, and a optional callback for exception handling?
Would also make consumer code cleaner and easier to reason about, as you don't need to distinguish between calls for exception handling and calls for data.

src/Solnet.Rpc/SolanaRpcBatchComposer.cs
```diff
@@ -0,0 +1,384 @@
...
+        }
+
+        /// <summary>
+        /// Process a failed batch response
+        /// </summary>
+        /// <param name="response">The failed batch RPC response</param>
+        /// <returns></returns>
+        public JsonRpcBatchResponse ProcessBatchFailure(RequestResult<JsonRpcBatchResponse> response)
+        {
+            // TODO - decide best approach - invoke each callback with exception?
```
tiago18c, line 221: I assume there is some error handling missing here?

src/Solnet.Rpc/SolanaRpcBatchComposer.cs
```diff
@@ -0,0 +1,384 @@
...
+                callback.Invoke(obj, ex);
+            };
+            return wrapper;
+        }
+
+        private static Action<JsonRpcBatchResponseItem, Exception> WrapTaskSource<T>(TaskCompletionSource<T> taskSource)
+        {
+            if (taskSource == null) return null;
+
+            // TODO - what about exceptions??
```
tiago18c, line 321: They're handled right? Remove comment.

src/Solnet.Rpc/SolanaRpcBatchWithAsyncs.cs
```diff
@@ -0,0 +1,145 @@
...
+using Solnet.Rpc.Types;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Solnet.Rpc
+{
+    public class SolanaRpcBatchWithAsyncs
```
tiago18c, line 13: Missing comments! Won't nitpick about private and internals, but public stuff must be properly commented for IDE tooltips, etc.

tiago18c: Methods bellow should be easy, just copy from IRpcClient

src/Solnet.Rpc/SolanaRpcBatchWithCallbacks.cs
```diff
@@ -0,0 +1,145 @@
...
+using Solnet.Rpc.Types;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Solnet.Rpc
+{
+    public class SolanaRpcBatchWithCallbacks
```
tiago18c, line 13: Comments from SolanaRpcBatchWithAsyncs apply here too!

src/Solnet.Rpc/SolanaRpcBatchWithAsyncs.cs
tiago18c: On a second note, this is the class we're removing right? If so, then ignore these comments, but do apply them on the callback version!

src/Solnet.Rpc/SolanaRpcClient.cs
```diff
@@ -1056,5 +1056,17 @@ public RequestResult<ulong> GetMinimumLedgerSlot()
                 .Result;
 
         #endregion
+
+        #region Batch
+
+        public async Task<RequestResult<JsonRpcBatchResponse>> SendBatchRequestAsync(JsonRpcBatchRequest reqs)
```
tiago18c, line 1062: Missing comment!

tiago18c (changes requested):
Overall looks very good!
Key points:
* Need to finish the comments for public methods
* Remove the unused batch mechanism
* Remove TODOs
* Add logging where needed (I think the trick here is going to be the IDs for events? Maybe set as the ID of the last or first request in a batch?)

Let me know if you need any assistance with this!

src/Solnet.Rpc/SolanaRpcClient.cs
liquizard: This functionality is provided via the `JsonRpcClient` base class as this was just a forwarding method and was incorrectly named. I've removed this method from `JsonRpcClient` and renamed the `SendBatchRequest` on the `JsonRpcClient` base class to `SendBatchRequestAsync`.

src/Solnet.Rpc/SolanaRpcBatchWithAsyncs.cs
liquizard: Class removed.

src/Solnet.Rpc/SolanaRpcBatchComposer.cs
liquizard: Good call.
Changing this, but it means a slight adjustment to the the lifecycle of a `SolanaRpcBatchComposer` as it will require a instance of `SolanaRpcClient` on construction to be able to share the `IdGenerator` to compose batches.

liquizard: This rabbit hole required exposing ID generation functionality via `IRpcClient` interface, added `GetNextIdForReq` function for this.

src/Solnet.Rpc/SolanaRpcBatchComposer.cs
liquizard: Removed and method changed to be internal as calling logic and flow deals with the success/failure outcome.

src/Solnet.Rpc/Core/Http/JsonRpcClient.cs
```diff
@@ -157,5 +157,106 @@ private async Task<RequestResult<T>> HandleResult<T>(JsonRpcRequest req, HttpRes
...
+            var requestsJson = JsonSerializer.Serialize(reqs, _serializerOptions);
+
+            try
+            {
+                // TODO - logging
+
+                // create byte buffer to avoid charset=utf-8 in content-type header
+                // as this is rejected by some RPC nodes
+                var buffer = Encoding.UTF8.GetBytes(requestsJson);
+                using var httpReq = new HttpRequestMessage(HttpMethod.Post, "/")
```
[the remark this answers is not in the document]
liquizard: Fixed.

src/Solnet.Rpc/SolanaRpcBatchComposer.cs
liquizard: Comment removed - implemented notification of batch failure to all callbacks of batch failure via `BatchRequestException` exception.

src/Solnet.Rpc/SolanaRpcBatchComposer.cs
liquizard: Yes - now covered by `SolanaRpcClientBatchTests.TestBatchFailed` test case.
Comment removed.

src/Solnet.Rpc/SolanaRpcBatchComposer.cs
liquizard: I can see that some developers may want separate success/failure callbacks and depending on the complexity of the calling code. 

This style with combined `(data, ex)` args has been used by several other developers including myself and nobody came back with this suggestion when asking for feedback. I suggest we ship what we have, and if the wider communitity would like to experiement with this separate callback style we can introduce a new `SolanaRpcBatchWithDualCallbacks` as an RFC like we tried and ultimately killed off with the `SolanaRpcBatchWithAsyncs` experiements.

src/Solnet.Rpc/Core/Http/JsonRpcClient.cs
liquizard: Please could you give me a steer on the best thing to use to construct the `EventId` with?
The batch contains multiple ids (for each inner request) but there isn't an overall id used for the batch itself.

Would it be ok to construct a composite id that includes the range? e.g. 4...23 - this may be inperfect as there could be gaps but is perhaps less crap than just choosing the first id.

src/Solnet.Rpc/Core/Http/JsonRpcClient.cs
liquizard: Gone with minimum request id as the id and `[batch on n]` as the event method.

src/Solnet.Rpc/Core/Http/JsonRpcClient.cs
liquizard: Gone with minimum request id as the id and `[batch on n]` as the event method.
Let me know if you'd like this changing...

src/Solnet.Rpc/Core/Http/JsonRpcClient.cs
liquizard: Gone with minimum request id as the id and `[batch on n]` as the event method.
Let me know if you'd like this changing...

src/Solnet.Rpc/Core/Http/JsonRpcClient.cs
```diff
@@ -157,5 +157,106 @@ private async Task<RequestResult<T>> HandleResult<T>(JsonRpcRequest req, HttpRes
...
+        /// </summary>
+        /// <param name="reqs">The message request.</param>
+        /// <returns>A task that represents the asynchronous operation that holds the request result.</returns>
+        public async Task<RequestResult<JsonRpcBatchResponse>> SendBatchRequest(JsonRpcBatchRequest reqs)
+        {
+            var requestsJson = JsonSerializer.Serialize(reqs, _serializerOptions);
+
+            try
+            {
+                // TODO - logging
```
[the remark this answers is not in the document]
liquizard: Gone with minimum request id as the id and `[batch on n]` as the event method.
Let me know if you'd like this changing...

tiago18c (approved):
It looks good!
Just one remaining issue: theres RateLimiter stuff in here, so its going to conflict with the other PR!

PR #244, opened 2021-10-27
Token wallet enhancements

liquizard:
| Status  | Type  | ⚠️ Core Change | 
| :---: | :---: | :---: | 
| Ready | Mini-Feature | Yes (addition) |

## Problem
Added a simple Sol property to TokenWallet to get native SOL balance.

```csharp
        public decimal Sol
        {
            get
            {
                const ulong LAMPORTS_PER_SOL = 1000000000;
                return Decimal.Round((decimal)this.Lamports / (decimal)LAMPORTS_PER_SOL,9);
            }
        }
```

## Solution
Added a very localised LAMPORTS_PER_SOL const for the conversion. This could do with a better home with wider cross-project scope. Is there (could/should there be) a Solnet-wide home consts like this?

### DIFF
--- a/SharedBuildProperties.props
+++ b/SharedBuildProperties.props
@@ -2,7 +2,7 @@
   xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <Product>Solnet</Product>
-    <Version>0.4.12</Version>
+    <Version>0.4.13</Version>
     <Copyright>Copyright 2021 &#169; Solnet</Copyright>
     <Authors>blockmountain</Authors>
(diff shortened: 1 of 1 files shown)


liquizard:
Yes. Great suggestions everyone, thank you.
Those are implemented now. 
Merge when you're happy, or give us a nod and I'll merge. 

"There are only two hard things in Computer Science: cache invalidation and naming things." [Phil Karlton]
`SolHelper` is a good name, thanks @Liebeck

liquizard:
Yikes - I broke the build. Sorry boys, I'll get the 🍻 in once I've fixed it.

Issue #109, opened 2021-06-21
Implement the Stake Program


murlokito:
Implement the Stake Program, which is a native program.

Documentation at:

[Solana Docs](https://docs.solana.com/developing/runtime-facilities/programs#stake-program)
[Rust Crate](https://docs.rs/solana-sdk/1.7.2/solana_sdk/stake/instruction/enum.StakeInstruction.html)
[Rust Source](https://docs.rs/solana-program/1.7.2/src/solana_program/stake/instruction.rs.html)

murlokito:
Closed by #251

PR #251, opened 2021-11-08
Stake Program

martinwhitman:
> ⚠️ NOTE: The AuthorizeWithStake is giving me fits still, but I think it's close; obv much cleanup/documenting to do, but I wanted to confirm this is what we're looking for

| In Progress | Feature | Yes (addition) | [Link](https://github.com/bmresearch/Solnet/issues/109) |

## Problem

Implementing the Stake Program from the SDK


martinwhitman:
Not sure what I did to break that unit test, if anything, but I'll figure that out

src/Solnet.Programs/SystemProgramData.cs
```diff
@@ -1,3 +1,4 @@
+using Newtonsoft.Json;
```
murlokito, line 1: Is this really necessary? We've been trying to minimize dependencies as much as possible

src/Solnet.Programs/Utilities/Serialization.cs
```diff
@@ -216,13 +216,23 @@ public static void WriteSingle(this byte[] data, float value, int offset)
...
+            return encoded;
+        }
         public static byte[] EncodeRustString(string data)
         {
-            byte[] stringBytes = Encoding.ASCII.GetBytes(data);
-            byte[] encoded = new byte[stringBytes.Length + sizeof(uint)];
+            byte[] stringBytes = Encoding.UTF8.GetBytes(data);
+            byte[] encoded = new byte[stringBytes.Length+sizeof(ulong)];
+
+            encoded.WriteU64((ulong)stringBytes.Length, 0);
```
murlokito, line 233: If I am not mistaken dynamic length strings have 4 bytes which are used to define the string length, this should still remain as a u32 or not?

src/Solnet.Rpc/Utilities/AddressExtensions.cs
```diff
@@ -19,6 +20,40 @@ public static class AddressExtensions
         /// The bytes of the `ProgramDerivedAddress` string.
         /// </summary>
         private static readonly byte[] ProgramDerivedAddressBytes = Encoding.UTF8.GetBytes("ProgramDerivedAddress");
+        private static byte[] EncodeRustString(string data)
```
murlokito, line 23: Can't the helper method in the utilities be used so we don't have this kind of code duplication?

murlokito: The same can be said for the functions below

src/Solnet.Programs/StakeProgram.cs
```diff
@@ -0,0 +1,365 @@
...
+        /// <summary>
+        /// The public key of the Recent Block Hashes System Variable. 
+        /// </summary>
+        public static readonly PublicKey
+            SysVarRecentBlockHashesKey = new("SysvarRecentB1ockHashes11111111111111111111");
+
+        /// <summary>
+        /// The public key of the Rent System Variable.
+        /// </summary>
+        public static readonly PublicKey SysVarRentKey = new("SysvarRent111111111111111111111111111111111");
```
murlokito, line 38: I would create a new static class for all of these sysvars

src/Solnet.Programs/StakeProgram.cs
```diff
@@ -0,0 +1,365 @@
...
+                AccountMeta.Writable(stake_pubkey, false),
+                AccountMeta.ReadOnly(authority_base, true),
+                AccountMeta.ReadOnly(SysVarClockKey, false),
+                AccountMeta.ReadOnly(new_authorized_pubkey, true)
+            };
+            if (custodian_pubkey != null)
+            {
+                keys.Add(AccountMeta.ReadOnly(custodian_pubkey, true));
+            }
+            AuthorizeCheckedWithSeedArgs authorizeWithSeed = new AuthorizeCheckedWithSeedArgs
```
murlokito, line 262: In general I have not done these ```SomeTypeArgs``` and have just passed the arguments directly to the method that encodes the data for the instruction. I would prefer if we kept this with a concise style all throughout the codebase, not having to do them for every program also makes the codebase leaner despite some of the function signatures being a little more verbose sometimes

src/Solnet.Programs/Models/Stake/State.cs
```diff
@@ -0,0 +1,48 @@
...
+    public class State
+    {
+        public struct Authorized
+        {
+            public PublicKey staker { get; set; }
+            public PublicKey withdrawer { get; set; }
+        }
+        public struct Delegation
+        {
+            public PublicKey voter_pubkey { get; set; }
```
murlokito, line 19: I think this will be my last comment, just regarding use of snake_case instead of PascalCase, again to maintain the code style all across the codebase.

murlokito (changes requested):
This looks good, I think we need to make sure about that rust string encoding wrt the length of the string being a u32 or u64 and the encoding because I think I used ASCII because of the name service, can't recall right now.

Other than that I think this is a matter of having consistent code style with the rest of the codebase (so as to not make it even harder for new contributors) and I think we should ship this

src/Solnet.Programs/SystemProgramData.cs
martinwhitman: Gone!

src/Solnet.Programs/StakeProgram.cs
martinwhitman: Moved SysVars to static class located at Solnet.Programs.SysVars

src/Solnet.Programs/StakeProgram.cs
martinwhitman: Eliminated

src/Solnet.Programs/Models/Stake/State.cs
martinwhitman: I think I got the spots where I did this.

src/Solnet.Rpc/Utilities/AddressExtensions.cs
martinwhitman: Eliminated that and the write functions below but kept createwithseed bc that doesn't exist elsewhere and that class is where other things from the pubkey.rs are implemented

martinwhitman: moved createwithseed to the utilities helper class to avoid circular dependency

src/Solnet.Programs/Utilities/Serialization.cs
martinwhitman: I think this might be an array vs vec issue

martinwhitman: Or rather, a vec vs string issue?

martinwhitman: https://doc.rust-lang.org/stable/std/string/struct.String.html#representation

It has to have length and capacity, so 2 x u32, so you're right that u32 is the length, and that explains why I could only get the tx to go through with a u64

martinwhitman: I can't get it to work without it being a u64

I am not sure but it seems like this is something like what is happening where it's looking for a u64

https://github.com/solana-labs/solana/blob/e540b1cf3c329c51de60a183529d0c39ee3d86d9/programs/bpf_loader/src/serialization.rs

martinwhitman:
I believe I have made the requested changes

murlokito (approved):
I only have two comments and other than these it looks very good to me:
- there's some comments missing in the StakeProgram class
- we're trying to reorganize Solnet.Programs before actually shipping v1.0 so it's less cluttered in the future, idea is to move towards a folder per program instead of having to cram everything into the `Models` folder, in this case just add a new `Stake` folder inside `Solnet.Programs` and move everything related in there

Once this is done I'll gladly merge this :)

martinwhitman:
I have made these changes. The TryCreateWithSeed is suddenly giving AddressWithSeedMismatch, though I am not sure what has changed to make it so. Fiddling with the EncodeRustString merely gives invalid transaction errors.

src/Solnet.Programs/Stake/StakeProgram.cs
```diff
@@ -0,0 +1,474 @@
...
+using System.Collections.Generic;
+using static Solnet.Programs.Models.Stake.State;
+
+namespace Solnet.Programs
+{
+    /// <summary>
+    /// Implements the Stake Program methods.
+    /// <remarks>
+    /// For more information see:
+    /// LINK HERE
```
tiago18c, line 14: Either add the links or remove the remarks. I'm assuming this would be links for the StakeProgram rust repos?

src/Solnet.Programs/Stake/StakeProgramData.cs
```diff
@@ -0,0 +1,277 @@
...
+            return data;
+        }
+        internal static byte[] EncodeAuthorizeData(PublicKey newAuthorizedPubkey, StakeAuthorize stakeAuthorize)
+        {
+            byte[] data = new byte[68];
+
+            data.WriteU32((uint)StakeProgramInstructions.Values.Authorize, MethodOffset);
+            data.WritePubKey(newAuthorizedPubkey, 4);
+            data.WriteU32((uint)stakeAuthorize, 36);
+            Console.WriteLine("Authorize data = "+ToReadableByteArray(data));
```
tiago18c, line 49: Please remove the `Console.WriteLine`s in the file.

src/Solnet.Rpc/Utilities/AddressExtensions.cs
```diff
@@ -94,7 +94,7 @@ public static bool TryFindProgramAddress(IEnumerable<byte[]> seeds, byte[] progr
         /// </summary>
         /// <param name="data">The data to hash.</param>
         /// <returns>The hash.</returns>
-        private static byte[] Sha256(byte[] data)
+        public static byte[] Sha256(byte[] data)
```
tiago18c, line 97: I think this should be moved into another class. Doesn't make sense to make this public here, but since its needed to be re-used, it makes sense to move it somewhere else. E.g. A `HashingUtilities` static class ?

src/Solnet.Rpc/Utilities/Utils.cs
```diff
@@ -0,0 +1,133 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Solnet.Rpc.Utilities
+{
+    public static class Utils
+    {
+
```
tiago18c, line 11: I can't find this being used anywhere, what am I missing?

test/Solnet.Programs.Test/Utilities/DeserializationUtilitiesTest.cs
```diff
@@ -271,7 +271,7 @@ public void TestReadRustStringException()
         public void TestReadRustString()
         {
             const string expected = "this is a test string"; // 21 chars
-            const int expectedLength = 25; // 21 chars + sizeof(uint) which is the encoding length padding
+            const int expectedLength = 29; // 21 chars + sizeof(u64) which is the encoding length padding
```
tiago18c, line 274: make it explicit `expectedLength = expected.Length + sizeof(ulong)` and there's no need for the comment.

But add a comment stating that it doesn't work for all characters due to utf-8 encoding.

src/Solnet.Examples/StakeExample.cs
```diff
@@ -0,0 +1,322 @@
...
+                    "dog1",
+                    3 * minbalanceforexception + 42,
+                    StakeProgram.StakeAccountDataSize,
+                    StakeProgram.ProgramIdKey))
+                .AddInstruction(SystemProgram.Transfer(
+                    a6,
+                    a5,
+                    5
+                    ))
+                //.AddInstruction(StakeProgram.Initialize(
```
tiago18c, line 295: Remove commented code. These examples need to be clean, so that someone new coming to Solnet has some examples to follow and to help understand things. Commented code won't help.

tiago18c (approved):
It looks good.
Consider it approved once you fix these small comments!

PS: thanks for the SysVars :)

murlokito (approved):
🚢🚢

tiago18c (approved):
LGTM ;)

Issue #252, opened 2021-11-11
[Bug] Bip39 Wallet generation is not this same as solana-keygen


Palkos83:
**Describe the bug**
I was trying to use SolNet 0.4.12 to generate some wallets which could be imported to Phantom Web Wallet.
However when generating wallets using the library the private keys from the store cannot be imported and are marked as invalid format. Wallets generated by solana-keygen are imported properly.

I tried using automatically generated mnemonics as well as replicating passphrase and word list from the solana-keygen.
The content of the json file generated by the   SecretKeyStoreService.SaveKeystore() is not this same as the one generated by solana-keygen (even when this same words and passphrase is used).

**To Reproduce**
Use solana-keygen to generate new wallet and save the json file.
Use this same passphrase and word list to generate new wallet using the library

**var mnem = "......";**   <- put your words list here
**var passPhrase  = "test" **<- replace with pass phrase used in the solana-keygen
**var wallet = new Wallet(mnem, WordList.English, passPhrase , SeedMode.Bip39);**

Save the wallet using secureKeyStore
_secretKeyStoreService.SaveKeystore(filePath, wallet);

Compare the two json files.

**Expected behavior**
The content of two generated json files should be this same and the private keys generated by the library should be importable to Phantom wallet.

**Desktop (please complete the following information):**
 - OS: Windows10
 - Browser: Chrome
 - Version 0.4.12

Please advise if I am missing something or using the library in invalid way.

murlokito:
Unless something changed in solana-keygen you should be using `SolanaKeyStoreService` and then importing the private key in Phantom, I think that might be what you're looking for? The `SecretKeyStoreService` implements the [web3 secret storage definition here](https://github.com/ethereum/wiki/wiki/Web3-Secret-Storage-Definition)

Palkos83:
Hi 

Many thanks for an answer - I am using the SolanaKeyStoreService. My secretKeyStoreSerivce is defined like this:

 **private readonly SolanaKeyStoreService _secretKeyStoreService = new ();**

Apologies if the name was a bit confusing.

Palkos83:
Hi guys... any update on this?

murlokito:
Hey @Palkos83, thanks for reporting this. We have been completely swamped so took me a little bit before I had time to dive into this. I have figured out the issue and will be patching this today + issuing a new release, stay tuned for 0.4.13 :)

murlokito:
[...]

PR #255, opened 2021-11-15
Fixing solana keystore

murlokito:
| Ready | Hotfix | No | #252  |

## Problem

Solana keystore was not saving the private key byte array as intended

## Solution

Changed the byte array it saves to the actual private key and not the mnemonic seed


tiago18c (approved):
LFG;

PR #262, opened 2021-11-22
Optimizes GetSpan and GetBytes due to memory allocation

### LINKED ISSUE #261: [Bug] Deserialization GetSpan copies serve no purpose but to create huge memory allocation
**Describe the bug**
In Deserialization.cs
public static Span<byte> GetSpan(this ReadOnlySpan<byte> data, int offset, int length)
    {
      if (offset + length > data.Length)
        throw new ArgumentOutOfRangeException(nameof (offset));
      byte[] numArray = new byte[length];
      data.Slice(offset, length).CopyTo((Span<byte>) numArray);
      return (Span<byte>) numArray;
    }
    
It seems this can be simplified in all paths to just data.Slice(offset,length) (and changing ReadOnlySpan<byte> to just Span<byte>). The copy doesn't serve any purpose. 

In my usage (subscribing to orderbookside updates), I see 150mb of allocation from this method in one minute of the program running.

murlokito:
| Ready| Refactor| No | Closes #261 |

## Problem

Memory allocation issues during deseralization using `GetSpan` and `GetBytes`

## Solution

Refactored `GetSpan` to not allocate buffers and just rely on slicing
Refactored `GetBytes` for the same reason and just use `.ToArray()`

### DIFF
--- a/src/Solnet.Programs/Utilities/Deserialization.cs
+++ b/src/Solnet.Programs/Utilities/Deserialization.cs
@@ -132,13 +132,11 @@ namespace Solnet.Programs.Utilities
         /// <param name="length">The desired length for the new span.</param>
         /// <returns>A <see cref="Span{T}"/> of bytes.</returns>
         /// <exception cref="ArgumentOutOfRangeException">Thrown when the offset is too big for the span.</exception>
-        public static Span<byte> GetSpan(this ReadOnlySpan<byte> data, int offset, int length)
+        public static ReadOnlySpan<byte> GetSpan(this ReadOnlySpan<byte> data, int offset, int length)
         {
             if (offset + length > data.Length)
                 throw new ArgumentOutOfRangeException(nameof(offset));
-            byte[] buffer = new byte[length];
-            data.Slice(offset, length).CopyTo(buffer);
-            return buffer;
+            return data.Slice(offset, length);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


tiago18c (approved):
LGTM;

Liebeck:
@murlokito I'm using some code out of https://github.com/MonkeDAO/observer/ which is build against 0.4.3
Yesterday, I updated the observer to use use 0.4.16 and I experienced that code that crawls collections did not work anymore.

In particular
`var metadataAccounts = await GetMetadataAccounts(69, _name);`

When iterating over the result of these metadataAccounts for an collection with
`var metadata = MetadataAccount.Deserialize(Convert.FromBase64String(metadataAccount.Account.Data[0]));`

I receive ArgumentOutOfRangeExceptions for https://github.com/bmresearch/Solnet/pull/262/commits/8d61d4925046f7c78bc3af111eea4f6422f813f6#diff-4bbbd5a885ee94a83d296fa7734815ddaed24d3d80d304f1c15ca9b3a179ef3fR137 Line 137.

I was only able to crawl collections again by reverting the commit from this pull request and compiling everything again. I didn't change anything else. I tried to debug for a couple of hours and was only able to trace it back to this commit.

Can you take a look if I'm going in the right direction or if this GetMetadataAccount might be affected by the change to ReadOnlySpan? I'm a little bit stuck at finding out why I'm having this problem after updating everything to 0.4.16

CC: @tiago18c

murlokito:
@Liebeck Was that issue coming up on all collections or did you just try scraping one collection? I find it hard to believe the problem was exactly that one change, by the looks of the structures it's probably blowing up reading the strings inside the `Data` structure, could you refactor [these](https://github.com/MonkeDAO/observer/blob/master/Observer/Models/Data.cs#L53) to use `GetString(offset, out string ...)` instead and see if this issue persists?

Liebeck:
@murlokito  I tried 4 different collections and I was unable to crawl any of them. That also included collections that I was able to parse before upgrading.

Then length parameter was very long, like 1032041 values or something in that range.

I'll try to use GetString later today, thanks!

PR #266, opened 2021-12-05
Deserialize TokenSwap State Account

thesoftwarejedi:
| Ready| Feature | No | _none_ |

## Problem

The `TokenSwap` state accounts are not able to be deserialized

## Solution

- Created `TokenSwapAccount` and a `Deserialize` method on it. 
- Created a base `Program` account with common program things (id and name for now).  

I realize that most programs are all static, but by making them objects it'd be possible to add other functionality in the future.  For instance, using a non-standard program id (mainnet vs testnet), and a common program registry that knows how to decode any account.  I have a working version of that in another branch. For now, prior to further discussion, just pushing like this to get the deserialize in there.

### DIFF
--- a/src/Solnet.Examples/TokenSwapExample.cs
+++ b/src/Solnet.Examples/TokenSwapExample.cs
@@ -22,6 +22,12 @@ namespace Solnet.Examples
 
         public void Run()
         {
+            //how to load current state
+            IRpcClient mainnetRpc = ClientFactory.GetClient(Cluster.MainNet);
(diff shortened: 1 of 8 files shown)


murlokito (approved):
This looks good to me. I like the base program addition due to the possibility of using programs on different clusters, which comes along very well with something I was already attempting to do with token lending

tiago18c:
LGTM :)

Regarding other accounts, I was thinking we could do a AccountDecoder similar to that of InstructionDecoder, specially for Anchor given that all accounts have a signature that should be able to match the type. I'm just struggling trying to picture how would this be useful. Maybe in an explorer kind of setting?

PR #285, opened 2021-12-31
Fixing json parse error

mcanerizci:
| Ready | Bug | No | - |

## Problem

It throws a "JsonException" because it can not parse the json when the CustomError value is more than the maximum value of int.

It is possible to see this situation in dozens of slot, but you can see it in slot number 113766234 as an example.

image

## Solution

Changed CustomError's type from int to long.

### DIFF
--- a/src/Solnet.Rpc/Converters/TransactionErrorJsonConverter.cs
+++ b/src/Solnet.Rpc/Converters/TransactionErrorJsonConverter.cs
@@ -105,7 +105,7 @@ namespace Solnet.Rpc.Converters
 
             if (reader.TokenType == JsonTokenType.Number)
             {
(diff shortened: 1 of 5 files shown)


tiago18c:
Nice find. But to make sure we're doing this 100% accurately, can you change to uint instead?
The Custom variant takes a u32 argument. 

Source: https://github.com/solana-labs/solana/blob/94a9b712b6985c8c0b7a55f824a394e35122de20/sdk/program/src/instruction.rs#L145

mcanerizci:
Yes you are right, uint is more convenient

tiago18c:
Thank you! Gonna merge. Expect a new version later

PR #287, opened 2022-01-04
Implements AddSignature method on TransactionBuilder

### LINKED ISSUE #192: Implement `AddSignature` on `TransactionBuilder`
In the context of Solnet.Serum we are, once again, avoiding users of the library having to pass in `Account` and/or `Wallet` to classes and relying on trusting the library is doing what it claims to do. Due to this, the current implementation and first official release of Solnet.Serum (v1.0.0) is going to be a bit heavier on performance than we'd like it to be. This is due to a few issues:

- When using the `TransactionBuilder` to craft a transaction and requesting another component to sign it, by using `CompileMessage()` the library user will be forced to decode the message and populating it with the appropriate signature(s) by doing the following:
```
Transaction tx = Transaction.Populate(Message.Deserialize(msgBytes), signatures);
byte[] txBytes = tx.Serialize(); 
```
- `TransactionBuilder` does not have an `AddSignature(byte[] signature)` method which could be used to build the transaction with

Proposed solution is the addition/rework of the following methods:

- `AddSignature(byte[] signature)` adds a signature to an internal list of signatures in the `TransactionBuilder`
- `Serialize()` which would serialize the transaction into wire format by using the signatures added with the previous method
- Reworking the internal logic so that this is completely compatible with the `Build(Account account)` and `Build(IList<Account> accounts)` methods

This would allow for even more flexibility when integrating with external components and APIs while also making this logic much more economic in terms of memory and performance.

RFC @tiago18c @liquizard @skynetcapital

murlokito:
| Ready| Feature | Yes| Closes #192  |

## Problem

See #192 

## Solution

Implemented `AddSignature` on `TransactionBuilder

Changed `Serialize` modifier to `public` and added a check on a case that occurs when using `AddSignature`

### DIFF
--- a/src/Solnet.Rpc/Builders/TransactionBuilder.cs
+++ b/src/Solnet.Rpc/Builders/TransactionBuilder.cs
@@ -45,9 +45,11 @@ namespace Solnet.Rpc.Builders
         /// <summary>
         /// Serializes the message into a byte array.
         /// </summary>
-        private byte[] Serialize()
+        public byte[] Serialize()
         {
             byte[] signaturesLength = ShortVectorEncoding.EncodeLength(_signatures.Count);
+            if (_serializedMessage == null)
+                _serializedMessage = _messageBuilder.Build();
             MemoryStream buffer = new(signaturesLength.Length + _signatures.Count * SignatureLength + _serializedMessage.Length);
 
             buffer.Write(signaturesLength);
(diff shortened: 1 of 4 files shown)


src/Solnet.Rpc/Builders/TransactionBuilder.cs
```diff
@@ -45,9 +45,11 @@ public TransactionBuilder()
         /// <summary>
         /// Serializes the message into a byte array.
         /// </summary>
-        private byte[] Serialize()
+        public byte[] Serialize()
         {
             byte[] signaturesLength = ShortVectorEncoding.EncodeLength(_signatures.Count);
+            if (_serializedMessage == null)
```
tiago18c, line 51: can we add a parameter to possibly force either way?
rationale: given tx size constraints, this would allow one to figure optimal TX contents during runtime for variable-length instruction data

tiago18c (approved):
Looks good to go, but discuss comment!

PR #297, opened 2022-01-17
Use stack instead of heap allocation in some parts of the code to imp…

Neo-vortex:
## Problem

Unnecessary heap allocation in some parts of the code

## Solution

Using  `stackalloc` to avoid heap allocation


tiago18c:
Please revert the format changes. We will take care of it in a later PR before 1.0. 
After that I will merge the stackallocs :)

PR #298, opened 2022-01-18
Use stack instead of heap allocation in some parts of the code to imp…

Neo-vortex:
Problem
Unnecessary heap allocation in some parts of the code

Solution
Using `stackalloc` to avoid heap allocation


tiago18c:
Just one last thing, `CryptoScrypt` allocs a ton of memory, and can't use stackalloc, that's why the tests are throwing `StackOverflowException`.

Neo-vortex:
I think using conditional stack allocation is a good way to avoid `stackoverflow` in this case

murlokito (approved):
This looks good to me!

src/Solnet.Rpc/Utilities/ShortVectorEncoding.cs
```diff
@@ -41,9 +41,8 @@ internal static byte[] EncodeLength(int len)
                 output[cursor] = (byte)elem;
                 cursor += 1;
             }
-
             byte[] bytes = new byte[cursor + 1];
-            Array.Copy(output, 0, bytes, 0, cursor + 1);
+            Array.Copy(output.ToArray(), 0, bytes, 0, cursor + 1);
```
tiago18c, line 45: Given that ToArray creates a new copy in the heap, use `output.CopyTo(output)` to prevent this.

src/Solnet.Wallet/Utilities/Base58Encoder.cs
```diff
@@ -163,12 +163,12 @@ public override byte[] DecodeData(string encoded)
             // Skip leading zeroes in b256.
             int it2 = size - length;
             // Copy result into output vector.
-            byte[] vch = new byte[zeroes + size - it2];
-            Array.Fill<byte>(vch, 0, 0, zeroes);
+            Span<byte> vch = stackalloc byte[zeroes + size - it2];
```
tiago18c, line 166: Same thing here, the array is returned, so a new copy is created in the heap. Revert to prevent more CPU usage in the copy and pointless stack allocation.

src/Solnet.Wallet/Utilities/BitWriter.cs
```diff
@@ -75,15 +75,15 @@ private static byte[] ToByteArray(BitArray bits)
...
+            Span<byte> array = stackalloc byte[arrayLength];
 
             for (int i = 0; i < bits.Length; i++)
             {
                 int b = i / 8;
                 int offset = i % 8;
                 array[b] |= bits.Get(i) ? (byte)(1 << offset) : (byte)0;
             }
-            return array;
+            return array.ToArray();
```
tiago18c, line 86: Same thing here, the array is returned, so a new copy is created in the heap. Revert to prevent more CPU usage in the copy and pointless stack allocation.

src/Solnet.Wallet/Utilities/BitWriter.cs
```diff
@@ -122,7 +122,7 @@ private static byte[] SwapEndianBytes(IReadOnlyList<byte> bytes)
                 }
                 output[i] = newByte;
             }
-            return output;
+            return output.ToArray();
```
tiago18c, line 125: Same thing here, the array is returned, so a new copy is created in the heap. Revert to prevent more CPU usage in the copy and pointless stack allocation.

src/Solnet.KeyStore/Crypto/Scrypt.cs
```diff
@@ -37,7 +41,7 @@ public static unsafe byte[] CryptoScrypt(byte[] password, byte[] salt, int n, in
             }
 
             /* 5: DK <-- PBKDF2(P, B, 1, dkLen) */
-            return SingleIterationPbkdf2(password, ba, dkLen);
+            return SingleIterationPbkdf2(password, ba.ToArray(), dkLen);
```
tiago18c, line 44: ba is being returned here, so revert the stack allocation for this one!

src/Solnet.KeyStore/Crypto/RandomBytesGenerator.cs
```diff
@@ -18,9 +19,9 @@ public byte[] GenerateRandomSalt()
 
         private static byte[] GenerateRandomBytes(int size)
         {
-            var bytes = new byte[size];
+            Span<byte> bytes = stackalloc byte[size];
             Random.NextBytes(bytes);
-            return bytes;
+            return bytes.ToArray();
```
tiago18c, line 24: Same thing here, the array is returned, so a new copy is created in the heap. Revert to prevent more CPU usage in the copy and pointless stack allocation.

tiago18c (changes requested):
Just some remarks left. Will merge after!

Neo-vortex:
To fully take advantage of `stack allocation` some other parts will need to change as well to accept `Span<byte>` instead of `byte[]`.
Closing this at this point.
probably will do it again when reach v 1.0.0 milestone

PR #300, opened 2022-01-18
Switch from Org.BouncyCastle.Crypto.Digests to System.Security.Crypto…

Neo-vortex:
## Problem

`SHA256` implementation in `System.Security.Cryptography `is about X2 times faster than` Org.BouncyCastle.Crypto.Digests `counterpart 

## Solution

Switch to ` System.Security.Cryptography` from `Org.BouncyCastle.Crypto.Digests` for `SHA256` for better performance

### DIFF
--- a/src/Solnet.Wallet/Utilities/Utils.cs
+++ b/src/Solnet.Wallet/Utilities/Utils.cs
@@ -1,7 +1,7 @@
 using Chaos.NaCl;
-using Org.BouncyCastle.Crypto.Digests;
 using System;
 using System.Collections.Generic;
+using System.Security.Cryptography;
 
 namespace Solnet.Wallet.Utilities
 {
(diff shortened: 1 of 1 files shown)


tiago18c:
Is there a use case where this becomes a visible problem?

Initially Solnet was using `System.Security.Crypto` but this isn't compatible on browser platforms (i.e. Blazor Wasm), so we changed into BouncyCastle for this single purpose.

Neo-vortex:
I did not face any problem in real life cases but benchmarks are defiantly faster with `System.Security.Crypto` 
I also checked Blazor Wasm with no problem.
here is the code I used and it performed well:
```csharp
@page "/counter"
@using System.Security.Cryptography

<PageTitle>Counter</PageTitle>

<h1>Counter</h1>

<p role="status">Current count: @currentCount</p>

<button class="btn btn-primary" @onclick="IncrementCount">Click me</button>

@code {
    private string currentCount = "";

    private void IncrementCount()
    {
        var data = new byte[6]{0, 1, 2, 3, 4, 5};
        var SHA256CHECKSUM = SHA256.Create();
        currentCount = Convert.ToBase64String( SHA256CHECKSUM.ComputeHash(data.AsSpan(0,6).ToArray()));
    }
}
```


I use `dotnet 6.0.101`
on `Chrome Version 97.0.4692.71 (Official Build) (64-bit)`
with latest  `PopOS`

tiago18c:
I double checked the docs, it seems the issue were the HMAC and key generation algorithms. Going to merge, thanks :)

PR #304, opened 2022-01-22
Fixed Authority issue with token mint deserialization.

### LINKED ISSUE #303: Serialization.TryCreateWithSeed broken
**Describe the bug**
Serialization.TryCreateWithSeed seems to be broken with Solana version 1.9.4 since it will not calculate the new key correct.

Right now you will end up in below validation.
https://github.com/solana-labs/solana/blob/ae955403878cbb3ab01743fef327d5dd1d67c5ab/runtime/src/system_instruction_processor.rs#L50

If I change `memoryStream.Write(EncodeRustString(seed));` into `memoryStream.Write(Encoding.UTF8.GetBytes(seed));` the correct key is created.

**Expected behavior**
Matching seed keys with `SystemProgram.CreateAccountWithSeed()`

tiago18c:
| Ready | Bug | No | Closes #303  |

## Problem

Key creation from seed was broken. Issue with deserialization of token mints.

## Solution

Fixed key creation. Added unit test.


murlokito (approved):
LGTM

PR #306, opened 2022-01-24
Fixed NameServiceClient query on all records for a given address.

tiago18c:
| Ready| Bug | No | [Link](<Issue link here>) |

## Problem

GetAllNamesByOwnerAsync does not return reverse regular .sol domains.

## Solution

Fixed logic error on handling reverse .sol domains queried.


murlokito (approved):
LGTM

PR #309, opened 2022-01-29
RcpClient GetBalance overload with Publickey

Neo-vortex:
## Problem

`GetBalance` in `IRpcClient` does not have an overload with `PublicKey` from `Solnet.Wallet`

## Solution

Adding extra overloads to support getting balance with `PublicKey` from `Solnet.Wallet`


murlokito:
Hmm, what's the reasoning behind these overloads? `PublicKey` implements implicit conversion to `string` so that this is not necessary.

PR #313, opened 2022-02-01
added PublicKey overload to GetBalance()

Neo-vortex:
## Problem

`GetBalance()` method in `IRpcClient` only accepts string (aka `Publickey` key)

## Solution

Add an overload to `GetBalance()` with `PublicKey`


murlokito:
I commented in the other PR you opened with these changes..

What's the reasoning behind these overloads? `PublicKey` implements implicit conversion to `string` for cases like these.

PR #316, opened 2022-02-07
Changed string ser/des names to avoid confusion.

tiago18c:
| Ready/Hold | Feature/Bug/Tooling/Refactor/Hotfix | Yes/No | [Link](<Issue link here>) |

## Problem

String serialization and deserialization method names are ambiguous.

## Solution

Changed these method names to be more explicit in what they do.


murlokito (approved):
LGTM

tiago18c:
This will require new deploys of all dependent libraries that decode strings. Maybe we should bump to 0.6?

murlokito:
> This will require new deploys of all dependent libraries that decode strings. Maybe we should bump to 0.6?

Could work, master already has the `Get/Write/BigInt` fixes which are kind of a big deal so wouldn't bother me to move faster with this version

tiago18c:
Should we change the Bincode methods to have similar semantics as the others?

PR #317, opened 2022-02-08
Added more methods to batching and missing docs.

tiago18c:
| Ready | Feature | No | [Link](<Issue link here>) |

## Problem

Missing important request in batching mechanism.

## Solution

Added missing public method docs.
Added the following methods to the batching mechanism:
* getSignaturesForAddress
* getAccountInfo
* getTokenLargestAccounts

### DIFF
--- a/src/Solnet.Rpc/SolanaRpcBatchWithCallbacks.cs
+++ b/src/Solnet.Rpc/SolanaRpcBatchWithCallbacks.cs
@@ -58,8 +58,15 @@ namespace Solnet.Rpc
 
         #region RPC Methods
 
-        // this is a sample set of methods with different response types
-
+        /// <summary>
+        /// Gets the balance for a certain public key.
+        /// <remarks>
+        /// The <c>commitment</c> parameter is optional, the default value <see cref="Commitment.Finalized"/> is not sent.
+        /// </remarks>
+        /// </summary>
+        /// <param name="pubKey">The public key.</param>
+        /// <param name="commitment">The state commitment to consider when querying the ledger state.</param>
+        /// <param name="callback">The callback to handle the result.</param>
         public void GetBalance(string pubKey, Commitment commitment = Commitment.Finalized,
                                Action<ResponseValue<ulong>, Exception> callback = null)
         {
@@ -67,6 +74,14 @@ namespace Solnet.Rpc
             _composer.AddRequest<ResponseValue<ulong>>("getBalance", parameters, callback);
         }
 
+        /// <summary>
+        /// Gets all SPL Token accounts by token owner.
+        /// </summary>
+        /// <param name="ownerPubKey">Public key of account owner query, as base-58 encoded string.</param>
+        /// <param name="tokenMintPubKey">Public key of the specific token Mint to limit accounts to, as base-58 encoded string.</param>
+        /// <param name="tokenProgramId">Public key of the Token program ID that owns the accounts, as base-58 encoded string.</param>
+        /// <param name="commitment">The state commitment to consider when querying the ledger state.</param>
+        /// <param name="callback">The callback to handle the result.</param>
         public void GetTokenAccountsByOwner(string ownerPubKey, string tokenMintPubKey = null,
                                             string tokenProgramId = null, Commitment commitment = Commitment.Finalized,
                                             Action<ResponseValue<List<TokenAccount>>, Exception> callback = null)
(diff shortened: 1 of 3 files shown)


src/Solnet.Rpc/SolanaRpcBatchWithCallbacks.cs
```diff
@@ -126,10 +196,23 @@ public void Flush()
                         KeyValue.Create("filters", filters),
                         HandleCommitment(commitment)));
 
-            _composer.AddRequest<List<AccountKeyPair>>("getProgramAccounts", parameters, callback);
```
liquizard, line 129: wow - so the type can be inferred by the compiler from the callback's type - that's very clever!

liquizard:
This looks great to me.

I like that you have been able to simplify the existing method innards by removing the generic type and allowing the compiler to figure out what it should be from the callback.

It occurred to me the other day that I did not implement `SendTransaction` which is kind of a big deal.
Whilst you have a PR open please could you add this too if you have a spare moment :-)

liquizard:
I think this should do it:

```csharp
        public void SendTransaction(byte[] transaction, 
                                    bool skipPreflight = false,
                                    Commitment preflightCommitment = Commitment.Finalized,
                                    Action<string, Exception> callback = null) =>
        
            SendTransaction(Convert.ToBase64String(transaction), skipPreflight, preflightCommitment, callback);

        
        public void SendTransaction(string transaction, 
                                    bool skipPreflight = false,
                                    Commitment preflightCommitment = Commitment.Finalized,
                                    Action<string, Exception> callback = null)
        {
            var parameters = Parameters.Create(
                  transaction,
                  ConfigObject.Create(
                        KeyValue.Create("skipPreflight", skipPreflight ? skipPreflight : null),
                        KeyValue.Create("preflightCommitment",
                            preflightCommitment == Commitment.Finalized ? null : preflightCommitment),
                        KeyValue.Create("encoding", BinaryEncoding.Base64)));

            _composer.AddRequest("sendTransaction", parameters, callback);
        }
```

tiago18c:
However, I think transactions its preferable to have more control with single transaction per request, its always good to have options when you need some sort of spam, or the instructions don't fit a single transaction.

liquizard:
I was thinking for an airdropper I might want to send multiple transactions in a single batch and then monitor their success.

liquizard:
If you can hang fire on merging I've couple more in the pipeline I'm working with... `GetMultipleAccounts` and `GetTokenAccountInfo`

murlokito (approved):
This looks good to me, maybe just add `SendTransaction` as well, even if it has limited use cases. Other than that, great job boys!

tiago18c:
The send TX was already done, but failed to commit 10mins ago 😅

liquizard:
Here we go, two more please:

```csharp
        public void GetMultipleAccounts(IList<string> accounts,
                                        Commitment commitment = Commitment.Finalized,
                                        Action<ResponseValue<List<AccountInfo>>, Exception> callback = null)
        {
            var parameters = Parameters.Create(
                    accounts,
                    ConfigObject.Create(
                        KeyValue.Create("encoding", "base64"),
                        HandleCommitment(commitment)));

            _composer.AddRequest("getMultipleAccounts", parameters, callback);

        }

        public void GetTokenAccountInfo(string pubKey,
                                        Commitment commitment = Commitment.Finalized,
                                        Action<ResponseValue<TokenAccountInfo>, Exception> callback = null)
        {
            var parameters = Parameters.Create(
                    pubKey,
                    ConfigObject.Create(
                        KeyValue.Create("encoding", "jsonParsed"),
                        HandleCommitment(commitment)));

            _composer.AddRequest("getAccountInfo", parameters, callback);

        }
```

Thanks for hanging on.

liquizard:
Re: internal token data decoded - this is a good case where you could have a flip to enable/disable this so you could isolate and quantify the performance improvement.

Is there an existing decoder function somewhere in `Solnet.Programs` you could point me to for decoding a token account's data?

tiago18c:
Compare the model above with these:

https://github.com/bmresearch/Solnet/blob/master/src/Solnet.Rpc/Models/TokenAccount.cs
https://github.com/bmresearch/Solnet/blob/master/src/Solnet.Rpc/Models/AccountData.cs


ps: just realized that the AccountData.cs file contains only 1 model related to accounts in general, and several ones related to tokens

tiago18c:
Yep! Imagine the compute cost from the validator, and even client side to parse all that json. Bet its much more efficient to just get the base64, parse the binary contents, and later provide fancy UI properties/methods.

tiago18c:
Also note, this isn't even a specific json rpc method. Just an abstraction we quickly added, given that you can only get all token accounts at once otherwise. But since we had the models already it cost nothing to add this.

Later on we added specific programs encoding/decoding.

liquizard:
Re: Imagine the compute cost...
I've just applied this approach to existing results from a call `getMultipleAccounts` in my app and managed to eliminate a whole wave of additional RPC requests. 
Well this was a very worthwhile conversation - thank you!

liquizard (approved):
A+ Would recommend! ⭐⭐⭐⭐⭐

liquizard:
Just out of interest, what's the difference between using your newly added `GetSignaturesForAddress` and the existing `GetConfirmedSignaturesForAddress2` ?

murlokito:
> Just out of interest, what's the difference between using your newly added `GetSignaturesForAddress` and the existing `GetConfirmedSignaturesForAddress2` ?

No difference AFAIK, just that `GetConfirmedSignaturesForAddress2` is deprecated and will be removed. I'm actually unsure if it wasn't already expected to have been removed since initially they wanted to get it off the API for v1.8, might just be a matter of time.

PR #322, opened 2022-02-11
Fixes WriteBorshString ignoring offset

### LINKED ISSUE #321: [Bug] WriteBorshString ignores offset param
**Describe the bug**
The Solnet.Programs.Utilities.Seralization.WriteBorshString method ignores the offset parameter when appending data to the byte array. This results in the overwriting of data at index 0. Note: this function has no usages, so it is minor

**To Reproduce**
- make a new byte array of length 8
- set first byte to an arbitrary value
- call WriteBorshString with a 3 character string and offset 1

**Expected behavior**
The arbitrary value at index 0 remains

**Desktop (please complete the following information):**
 - OS: Windows
 - Browser N/A
 - Version 0.6.0

murlokito:
| Ready| Hotfix | Yes | Closes #321 |

## Problem

`WriteBorshString` would ignore the offset parameter and overwrite at offset 0 of the destination buffer

## Solution

Used the correct offset for the underlying `WriteU32` and `WriteSpan` calls

### DIFF
--- a/src/Solnet.Programs/Utilities/Serialization.cs
+++ b/src/Solnet.Programs/Utilities/Serialization.cs
@@ -225,8 +225,8 @@ namespace Solnet.Programs.Utilities
             if(offset + sizeof(uint) + stringBytes.Length > data.Length)
                 throw new ArgumentOutOfRangeException(nameof(offset));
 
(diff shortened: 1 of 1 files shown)


tiago18c:
LGTM;

PR #333, opened 2022-02-19
Fix to BaseClient signing process.

tiago18c:
| Ready | Bug | Yes | [Link](<Issue link here>) |

## Problem

BaseClient does not sign transactions properly.

## Solution

Fixed BaseClient signing process.
Additional QoL changes in nonce return from PDA derivation.


murlokito (approved):
LGTM

PR #341, opened 2022-02-24
Extending base client

tiago18c:
| Ready| Feature | Yes | [Link](<Issue link here>) |

## Problem

Base client functionality is too limited for anchor features.

## Solution

Extending base client to help error handling.

### DIFF
--- a/src/Solnet.Programs/Abstract/BaseClient.cs
+++ b/src/Solnet.Programs/Abstract/BaseClient.cs
@@ -22,6 +22,11 @@ namespace Solnet.Programs.Abstract
     public abstract class BaseClient
     {
         /// <summary>
+        /// The program address.
+        /// </summary>
+        public PublicKey ProgramID { get; }
+
+        /// <summary>
         /// The RPC client.
         /// </summary>
         public IRpcClient RpcClient { get; init; }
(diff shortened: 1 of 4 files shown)


src/Solnet.Programs/Abstract/BaseClient.cs
```diff
@@ -21,6 +21,11 @@ namespace Solnet.Programs.Abstract
     /// </summary>
     public abstract class BaseClient
     {
+        /// <summary>
+        /// The program address.
+        /// </summary>
+        public PublicKey ProgramID { get; }
```
murlokito, line 27: Casing/Naming? [see](https://github.com/bmresearch/Solnet/blob/8a9b0272f27d29f06dc2026fdb6485502cdd1ea3/src/Solnet.Programs/Abstract/BaseProgram.cs#L21)

src/Solnet.Programs/Abstract/TransactionalBaseClient.cs
```diff
@@ -0,0 +1,159 @@
...
+            tb.AddInstruction(instruction);
+
+            var recentHash = await RpcClient.GetRecentBlockHashAsync();
+
+            tb.SetRecentBlockHash(recentHash.Result.Value.Blockhash);
+            tb.SetFeePayer(feePayer);
+
+            var wireFmt = tb.CompileMessage();
+
+            var msg = Message.Deserialize(wireFmt);
```
murlokito, line 67: Can mitigate these ser/des of the msg/tx wire format

```csharp
var wireFmt = tb.CompileMessage();

var feePayerSignature = signingCallback(wireFmt, feePayer);

tb.AddSignature(feePayerSignature);

...


return await RpcClient.SendTransactionAsync(tb.Serialize(), commitment: commitment);

```

murlokito (approved):
Just those two comments, otherwise it looks good.

Would prefer if the `ProgramID` adhered to the same convention of the property name that the `BaseProgram` has, the rest is just some performance polishing.

LGTM

PR #345, opened 2022-02-28
small performance improvement by using unsafe code

Neo-vortex:
>  

| Ready | Optimization | No | NO|

## Problem

A  tiny performance gain in serialization utility ( see [this](https://stackoverflow.com/questions/4980881/what-is-fastest-way-to-convert-bool-to-byte)).

## Solution

using `unsafe` code without explicit converting and also removing branching.

## Before & After Performance

```
.NET SDK=6.0.102
  [Host]     : .NET 6.0.2 (6.0.222.6406), X64 RyuJIT
  DefaultJob : .NET 6.0.2 (6.0.222.6406), X64 RyuJIT


|                    Method |     Mean |   Error |  StdDev | Allocated |
|-------------------------- |---------:|--------:|--------:|----------:|
| explicit_branching_assign | 504.0 ns | 2.01 ns | 1.78 ns |         - |
|             unsafe_assing | 205.7 ns | 0.99 ns | 0.92 ns |         - |
```
## Benchmark Code

```csharp


namespace Benchmark
{
    [MemoryDiagnoser]
    public  class Benchmark
    {
        private static byte _dummy;
        [Benchmark]
        public void explicit_branching_assign()
        {
            byte result = 0;
            var _the_bool = DateTimeOffset.Now.ToUnixTimeMilliseconds() % 2 ==0;
            for (var i = 0; i < 500; i++)
            {
                result = _the_bool ? (byte) (1) : (byte) 0;
            }
            _dummy = result ;
        }

        [Benchmark]
        public  unsafe void unsafe_assing(){
            byte result = 0;
            var _the_bool = DateTimeOffset.Now.ToUnixTimeMilliseconds() % 2 ==0;
            for (var i = 0; i < 500; i++)
            {
                result = *((byte*)(&_the_bool));
            }
            _dummy = result ;
        }
    }
    public static class MainClass
    {
        public static void Main()
        {
            var result = BenchmarkRunner.Run<Benchmark>();
        }
    }
}

```

### DIFF
--- a/SharedBuildProperties.props
+++ b/SharedBuildProperties.props
@@ -2,7 +2,7 @@
   xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <Product>Solnet</Product>
-    <Version>6.0.3</Version>
+    <Version>6.0.4</Version>
     <Copyright>Copyright 2022 &#169; Solnet</Copyright>
     <Authors>blockmountain</Authors>
     <PublisherName>blockmountain</PublisherName>
--- a/src/Solnet.Programs/Solnet.Programs.csproj
+++ b/src/Solnet.Programs/Solnet.Programs.csproj
@@ -2,6 +2,7 @@
 
   <PropertyGroup>
     <TargetFramework>net6.0</TargetFramework>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/Solnet.Programs/Utilities/Serialization.cs
+++ b/src/Solnet.Programs/Utilities/Serialization.cs
@@ -35,11 +35,11 @@ namespace Solnet.Programs.Utilities
         /// <param name="value">The boolean value to write.</param>
         /// <param name="offset">The offset at which to write the 8-bit unsigned integer.</param>
         /// <exception cref="ArgumentOutOfRangeException">Thrown when the offset is too big for the data array.</exception>
-        public static void WriteBool(this byte[] data, bool value, int offset)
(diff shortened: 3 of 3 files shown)


tiago18c:
The change looks good.
Just one minor thing: please revert the line ending changes.

ps: also, bump the version to 6.0.4 and I'll merge, the github actions will do the rest :)

Neo-vortex:
I am a bit confused here!
Files in Utilities namespace are LF ending, while some others like SystemPorgram.cs is CRLF
Now I changed it back to CRLF i think.
is it ok or I change it to LF ?

tiago18c:
The idea was to revert just this file in order to keep the PR simple. 
We wanted to do a global reformat in our v1, but we forgot about it.

Regarding the line endings, is there any way to force a specific line ending per project? Been looking into it, but so far I can't find anything. We're using windows, linux and macos and it would be great to have a way to force line endings to be equal across

Neo-vortex:
probably the best way would be to let everyone works with their own native line ending.but configure `git` to do `autocrlf`.
You can also configure `.gitattributes` like this:
```
*.cs text eol=crlf
```
to force line ending. but  the IDE on Mac and Linux should support CRLF.
Read more [here](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings)

tiago18c:
Just found out something:

`dotnet format` doesn't respect `.editorconfig` parameter `end_of_line` if run under Windows, but does using MacOS.

I'll open a new PR later to run `dotnet format` and force `crlf` on all files :)

Neo-vortex:
Good!
That's another way to do it!
Can you please merge this now?!

PR #350, opened 2022-03-11
Fixed bad indexing on names without reverse lookup account.

tiago18c:
| Ready| Bug | No | [Link](<Issue link here>) |

## Problem

`ArgumentOutOfRangeException` when calling `GetAllNamesByOwnerAsync` on the `NameServiceClient`.

## Solution

Fixed bad indexing when iterating over names to get the reverse name (and the  respective reverse lookup account doesn't exist)

### DIFF
--- a/SharedBuildProperties.props
+++ b/SharedBuildProperties.props
@@ -2,7 +2,7 @@
   xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
     <Product>Solnet</Product>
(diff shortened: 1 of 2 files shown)


murlokito (approved):
Looks good to me. You're gonna merge the public keys QOL stuff before this one right?

tiago18c:
> Looks good to me. You're gonna merge the public keys QOL stuff before this one right?

Yep! only want to do a single version bump.

PR #355, opened 2022-03-12
Bad request setup when trying to fetch all names owned by a given add…

tiago18c:
| Ready | Bug | No | [Link](<Issue link here>) | 

## Problem

Unable to fetch addresses owned by a given address that owns more than 100 addresses.

## Solution

Fixed call to getMultipleAccounts possibly requesting more than 100 accounts.


murlokito (approved):
LGTM

PR #403, opened 2022-06-30
PublicKey constructor will quietly accept invalid inputs

liquizard:
| Status  | Type  | ⚠️ Core Change | 
| :---: | :---: | :---: | 
| Ready | Bugfix | Yes | 

## Problem
Encountered the error `invalid transaction: Transaction failed to sanitize accounts offsets correctly` when submiting a Serum SettleFunds command. Managed to pinpoint the root cause of the problem being a rouge space on the end of a public key string constant. Removing this space resolved the issue - this must cause some internal side-effect near the `TransactionBuilder`.

Further investigation found that PublicKey constructor would silently accept all manner of shenanigans. 
Also found that `PublicKey.IsValid` implementation relies on `Base58Encoder` validation logic that would accept whitespace.

## Solution
Added a `FastCheck` internal method to `PublicKey` and added calls to it from constructor and `IsValid`.
Added and modified a few tests.

## Deploy Notes
This is potentially a breaking change to any client apps that accept public keys that include whitespace before or after and pass them directly to the constructor. On balance, I believe this enhancement will prevent more problems than it will cause.

### DIFF
--- a/src/Solnet.Wallet/PublicKey.cs
+++ b/src/Solnet.Wallet/PublicKey.cs
@@ -79,7 +79,9 @@ namespace Solnet.Wallet
         /// <param name="key">The public key as base58 encoded string.</param>
         public PublicKey(string key)
         {
-            Key = key ?? throw new ArgumentNullException(nameof(key));
+            if (key == null) throw new ArgumentNullException(nameof(key));
+            if (!FastCheck(key)) throw new ArgumentException("publickey contains a non-base58 character");
+            Key = key;
         }
 
         /// <summary>
@@ -201,10 +203,11 @@ namespace Solnet.Wallet
         /// <returns>Returns true if the input is a valid key, false otherwise.</returns>
         public static bool IsValid(string key, bool validateCurve = false)
         {
-            if(!string.IsNullOrEmpty(key))
+            if (!string.IsNullOrEmpty(key))
             {
                 try
                 {
+                    if (!FastCheck(key)) return false;
                     return IsValid(Encoders.Base58.DecodeData(key), validateCurve);
                 }
                 catch(Exception)
@@ -247,6 +250,18 @@ namespace Solnet.Wallet
         public static bool IsValid(ReadOnlySpan<byte> key, bool validateCurve = false)
         {
             return key != null && key.Length == PublicKeyLength && (!validateCurve || key.IsOnCurve());
+        }
+
+        /// <summary>
+        /// Fast validation to determine whether this is a valid public key input pattern. 
+        /// Checks are valid characters for base58 and no whitespace.
+        /// Avoids performing the conversion to a buffer and checking it is actually 32 bytes as a permformance trade-off.
+        /// </summary>
+        /// <param name="value">public key value to check</param>
+        /// <returns>true means good, false means bad</returns>
+        private static bool FastCheck(string value)
+        {
+            return Base58Encoder.IsValidWithoutWhitespace(value);
         }
 
         #region KeyDerivation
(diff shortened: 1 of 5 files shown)


src/Solnet.Wallet/PublicKey.cs
```diff
@@ -79,7 +79,9 @@ public PublicKey(byte[] key)
         /// <param name="key">The public key as base58 encoded string.</param>
         public PublicKey(string key)
         {
-            Key = key ?? throw new ArgumentNullException(nameof(key));
+            if (key == null) throw new ArgumentNullException(nameof(key));
+            if (!FastCheck(key)) throw new ArgumentException("publickey contains a non-base58 character");
```
tiago18c, line 83: Was thinking about adding a flag to check or ignore checking, with default to check?

```
public PublicKey(string key, bool validateKey = true)
{
            if (key == null) throw new ArgumentNullException(nameof(key));
            if (validateKey && !FastCheck(key)) throw new ArgumentException("publickey contains a non-base58 character");
 
 ....
}
```

Discuss.

src/Solnet.Wallet/Utilities/Base58Encoder.cs
```diff
@@ -170,5 +172,23 @@ public override byte[] DecodeData(string encoded)
...
+
+
+        /// <summary>
+        /// Strict validation with no whitespace allowed
+        /// </summary>
+        /// <param name="value">Base58 string data</param>
+        /// <returns></returns>
+        /// <exception cref="ArgumentNullException"></exception>
+        public static bool IsValidWithoutWhitespace(string value)
+        {
```
tiago18c, line 184: Please delete the method `IsMaybeEncoded ` that had same functionality but is now obsolete given yours should beat it in performance by several magnitudes due to lookup dictionary.

tiago18c (changes requested):
Discuss comments. Otherwise looks good!

PS: I traced the actual problem to the method 'FindAccountIndex' in the 'MessageBuilder` class. I'll issue a fix later to this one! This fix should also address performance!
(root problem is that decoder was correctly trimming whitespace, but the message builder is inefficiently converting between binary and b58 multiples times and comparing the original pk string with a converted one that loses the trailing space. )

src/Solnet.Wallet/PublicKey.cs
liquizard: Eliminating any possibilities that a `PublicKey` coudl have been instantiated with an invalid key prevents a whole category of bugs in all code paths that use `PublicKey` instances. Allowing unsafe keys will necessetate duplicate checks in library code that accepts `PublicKey` instances to double-check whether it the value actually valid.

I cannot think of a single use case where you would want an instance of `PublicKey` that contains an invalid key. If there are, then perhaps a `PublicKeyUnchecked` derived class would better and a more explicit choice for developers. Any library code can use the `PublicKey` and be safe in the knowledge all instances have been pre-validated.

src/Solnet.Wallet/Utilities/Base58Encoder.cs
liquizard: Done

src/Solnet.Wallet/PublicKey.cs
liquizard: image

tiago18c: You convinced me sir! A while ago we were doing decoding stuffs and we had to change the `PublicKey` class to prevent too much checks on the constructor (it was doing encoding back then). If perf is a problem, one could derive and use specifically yeah!

tiago18c:
Thanks for the change. Don't merge yet, let me merge the other change and we can do 2PRs in 1 version

PR #404, opened 2022-06-30
Fix to MessageBuilder.

tiago18c:
| Ready | Bug/Refactor | No | [Link](<Issue link here>) |

## Problem

Message builder would introduce bad account index without warning.

## Solution

Throwing Exceptions when the account index is not found. Fixed MessageBuilder performance - no longer encoding / decoding all keys between binary and b58.


murlokito (approved):
LGTM

PR #405, opened 2022-07-01
Account ordering now matches web3.js

### LINKED ISSUE #397: Restore compatibility with web3.js on partial signing
As discussed here #396

tiago18c:
| Ready | Bug/Refactor |  Closes #397 |

## Problem

Partial signing is incompatible with web3.js

## Solution

Fixed account ordering to match web3.js. 

Minor version bump instead of patch.


murlokito (approved):
LFG

Issue #426, opened 2022-12-19
[Bug] call to getTransaction on a versioned transaction fail


nazbrok:
Solana introduced [versioned transaction]( https://edge.docs.solana.com/developing/versioned-transactions)

currently Solnet support only LEGACY transaction because the optional parameter `maxSupportedTransactionVersion` is not set on getBlock and getTransaction

to my understanding we just need to always set this optional parameter to 0 to support both mode. 

**To Reproduce**

Metaplex action house started to use versioned transaction.
Here is a (failed) TX using versioned transaction : [X1GTDnoauLQvM7VoZjC492dVY1YRmYCbUDPX7WwyuiWizFyigJfdGJh61Qrkr7hjAqoo8aVWJLWXEeRgqv7JQPp](https://solscan.io/tx/X1GTDnoauLQvM7VoZjC492dVY1YRmYCbUDPX7WwyuiWizFyigJfdGJh61Qrkr7hjAqoo8aVWJLWXEeRgqv7JQPp)

calling getTransaction for this tx return this json

`{"jsonrpc":"2.0","error":{"code":-32015,"message":"Transaction version (0) is not supported by the requesting client. Please try the request again with the following configuration parameter: \"maxSupportedTransactionVersion\": 0"},"id":54}`

BifrostTitan:
https://github.com/bmresearch/Solnet/pull/419/commits/eac74306442d8585646c90cb17bf7b2b57f4a529

Added support for versioned transactions. I believe compile message needs a update to write versioned transaction from the client not totally sure so this only covers the reading portion.

nazbrok:
> [eac7430](https://github.com/bmresearch/Solnet/commit/eac74306442d8585646c90cb17bf7b2b57f4a529)
> 
> Added support for versioned transactions. I believe compile message needs a update to write versioned transaction from the client not totally sure so this only covers the reading portion.

indeed this PR only aim to fix the reading portion for the moment.
My knowledge of Solana isn't great so I didn't dive too much into the writing part (yet)

nazbrok:
Well looks like this PR only fix the call to getTransaction and getBlock and is finally not enough for the reading portion

The transaction format changed so Solnet currently fail to decode the tx because of the table lookup I think.

BifrostTitan:
https://github.com/bmresearch/Solnet/commit/eac74306442d8585646c90cb17bf7b2b57f4a529
[...]

PR #428, opened 2023-01-16
Add optionnal parameter maxSupportedTransactionVersion to support ver…

nazbrok:
This PR aim to fix the issue #426  

Solana added versionned transaction since EPOCH 358.

Current version of solnet only support `getTransaction` and `getBlock` on LEGACY version.
To be able to call getTransaction or getBlock on versionned transaction, the optionnal parameter `maxSupportedTransactionVersion` need to be set to `0`.

### DIFF
--- a/src/Solnet.Rpc/Models/MessageV0.cs
+++ b/src/Solnet.Rpc/Models/MessageV0.cs
@@ -1,0 +1,324 @@
+using Solnet.Rpc.Utilities;
+using Solnet.Wallet;
+using Solnet.Wallet.Utilities;
+using System;
+using System.Collections.Generic;
+using System.IO;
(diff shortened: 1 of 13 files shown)


src/Solnet.Rpc/Models/MessageV0.cs
```diff
@@ -0,0 +1,324 @@
...
+using Solnet.Wallet;
+using Solnet.Wallet.Utilities;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace Solnet.Rpc.Models
+{
+    public class VersionnedMessage
```
StephenDemanuele, line 11: typo: VersionedMessage

Issue #477, opened 2025-01-25
Add Address Lookup Table Support


kingsznhone:
**Is your feature request related to a problem? Please describe.**
address lookup table is useful while sending massive instruction

**Describe the solution you'd like**
is there any chance to add versioned transaction compile with address lookup table support?

BifrostTitan:
Address lookup table can be specified in the transaction builder for versioned transactions. Latest version of solnet is available on nuget for NET8.0

https://github.com/bmresearch/Solnet/blob/master/src/Solnet.Rpc/Models/VersionedTransaction.cs

kingsznhone:
But I didn't found  any extraction lookup table index operation for each instruction in TransactionBuilder. Versioned message only attach ALT's Pubkey and index in the end of byte array.

icaca:
Hello, how can I query ATL like js, then use ALT in the signature, and then compile it into V0Message.

js code such as:

```
// get address lookup table accounts
const addressLookupTableAccounts = await Promise.all(
transaction.message.addressTableLookups.map(async (lookup) => {
return new AddressLookupTableAccount({
key: lookup.accountKey,
state: AddressLookupTableAccount.deserialize(await connection.getAccountInfo(lookup.accountKey).then((res) => res.data)),
})
}))

transaction.message = message.compileToV0Message(addressLookupTableAccounts)
```


Reference: https://station.jup.ag/docs/old/additional-topics/composing-with-versioned-transaction

kingsznhone:
> Hello, how can I query ATL like js, and then compile it into V0Message.
> 
> js code such as:
> 
> ```
> // get address lookup table accounts
> const addressLookupTableAccounts = await Promise.all(
> transaction.message.addressTableLookups.map(async (lookup) => {
> return new AddressLookupTableAccount({
> key: lookup.accountKey,
> state: AddressLookupTableAccount.deserialize(await connection.getAccountInfo(lookup.accountKey).then((res) => res.data)),
> })
> }))
> 
> transaction.message = message.compileToV0Message(addressLookupTableAccounts)
> ```
> 
> Reference: https://station.jup.ag/docs/old/additional-topics/composing-with-versioned-transaction

sadly message builder do not implements ALT complie yet. you can't

icaca:
> > Hello, how can I query ATL like js, and then compile it into V0Message.
> > js code such as:
> > ```
> > // get address lookup table accounts
> > const addressLookupTableAccounts = await Promise.all(
[...]

PR #478, opened 2025-02-01
Add Address lookup table Instruction

kingsznhone:
>  

| Ready | Refactor | Yes | |

## Problem
#477 
Add address lookup table instruction support for futhur development.

Make some crypto refactor on modern runtime.

## Solution

1. Removed `Org.BouncyCastle` references in `Mnemonic.cs` and `Ed25519Bip32.cs` and used built-in encryption methods instead.
2. added `ImportMany` methods in `Account.cs`.
3. Optimized `LoadAsync` method in `HardcodedWordListSource.cs`.
4. Used `RandomNumberGenerator` in `RandomUtils.cs` instead, which is crypto safe on all platform.
5.  Added address lookup table program related files to support create, freeze, expand, deactivate and close instructions.

### DIFF
--- a/src/Solnet.Programs/AddressLookupTableProgram.cs
+++ b/src/Solnet.Programs/AddressLookupTableProgram.cs
@@ -1,0 +1,140 @@
+using System.Collections.Generic;
+using Solnet.Rpc.Models;
+using Solnet.Wallet;
+
+namespace Solnet.Programs
+{
+    public static class AddressLookupTableProgram
+    {
+        /// <summary>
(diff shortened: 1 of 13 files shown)


kingsznhone:
I've tried using AES-GCM to replace AES-CTR in `Solnet.KeyStore`, But Unit test partially failed due to `...KeyStore.json` in resource not updated. And I'm not fully understand what does KeyStore do. So just leave it alone for now.

BifrostTitan (approved):
Looks good. Will do some testing in a bit

PR #487, opened 2025-03-19
use asynchronous with RateLimiter

### LINKED ISSUE #486: Thread.Sleep terrible for Task
https://github.com/bmresearch/Solnet/blob/71d0a30bd811818eb08036a27d02215d1f4e7c71/src/Solnet.Rpc/Utilities/RateLimiter.cs#L62

Can we make it to async await with Task.Delay ? 

```csharp
        public async Task WaitFireAsync(CancellationToken cancellationToken = default)
        {

            var checkTime = DateTime.UtcNow;
            var resumeTime = NextFireAllowed(checkTime);
            var snoozeMs = resumeTime.Subtract(checkTime).TotalMilliseconds;
            while (DateTime.UtcNow <= resumeTime)
                await Task.Delay(50, cancellationToken);

            // record this trigger
            if (_duration_ms > 0)
                _hit_list.Enqueue(DateTime.UtcNow);

        }


    //use
    await (_rateLimiter?.WaitFireAsync() ?? Task.CompletedTask);
```
and set `[Obsolete]` on `Fire` for keep testing

tqk2811:
| Ready| Refactor| No | [Link](https://github.com/bmresearch/Solnet/issues/486) |

## Problem

Thread.Sleep is terrible for Task

## Solution

Change it to asynchronous with Task.Delay

## Before & After Screenshots

**BEFORE**:
image
image
image

**AFTER**:
image
image
image

### DIFF
--- a/src/Solnet.Rpc/Core/Http/JsonRpcClient.cs
+++ b/src/Solnet.Rpc/Core/Http/JsonRpcClient.cs
@@ -78,8 +78,8 @@ namespace Solnet.Rpc.Core.Http
             try
             {
                 // pre-flight check with rate limiter if set
-                _rateLimiter?.Fire(); 
-                
+                await (_rateLimiter?.WaitFireAsync() ?? Task.CompletedTask);
+
                 // logging
                 _logger?.LogInformation(new EventId(req.Id, req.Method), $"Sending request: {requestJson}");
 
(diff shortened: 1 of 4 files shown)


BifrostTitan:
Yeah should be using TAP and not blocking the thread using sleep.

PR #493, opened 2025-04-18
Refactor PrivateKey class to enforce immutability and improve key security

ManOguaR:
This PR replaces the original implementation of the `PrivateKey` class with a secure, immutable version that eliminates unsafe public exposure of sensitive cryptographic material.

### Summary of changes:
- Removed `DebuggerDisplay` and `ToString()` that previously exposed the Base58-encoded private key.
- Replaced mutable public properties (`Key`, `KeyBytes`) with a single immutable backing field and safe readonly access.
- Enforced strict immutability after construction.
- Lazy Base58 encoding with internal caching (`GetBase58()`), no longer exposed via `ToString()`.
- Updated `Equals()` to perform byte-level comparison rather than string-based.
- Replaced `GetHashCode()` with a fast, consistent implementation derived from internal key bytes.
- Signature logic remains intact (`Sign()` using `Ed25519.Sign`).

### Notes:
- **This breaks tests/examples** that relied on the insecure `.ToString()` or public access to `Key`. This is intentional.
- Key exposure through debugger/logging is now fully eliminated.
- No functional change to wallet creation, signing, or keypair loading.

This refactor lays groundwork for improving keystore handling, memory hygiene, and test coverage without leaking sensitive material.

Let me know if you'd like this split further or integrated into an upcoming security milestone.

### DIFF
--- a/src/Solnet.Examples/HelloWorldExample.cs
+++ b/src/Solnet.Examples/HelloWorldExample.cs
@@ -16,7 +16,7 @@ namespace Solnet.Examples
             Console.WriteLine("Hello World!");
             Console.WriteLine($"Mnemonic: {wallet.Mnemonic}");
             Console.WriteLine($"PubKey: {wallet.Account.PublicKey.Key}");
-            Console.WriteLine($"PrivateKey: {wallet.Account.PrivateKey.Key}");
+            //Console.WriteLine($"PrivateKey: {wallet.Account.PrivateKey.Key}");
 
             IRpcClient rpcClient = ClientFactory.GetClient(Cluster.TestNet);
 
--- a/src/Solnet.Examples/SolanaKeygenKeyGeneration.cs
+++ b/src/Solnet.Examples/SolanaKeygenKeyGeneration.cs
@@ -18,16 +18,16 @@ namespace Solnet.Examples
             var solKeygenWallet = new Wallet.Wallet(mnemonic, passphrase, SeedMode.Bip39);
 
             Console.WriteLine($"SOLLET publicKey>b58 {solKeygenWallet.Account.PublicKey}");
-            Console.WriteLine($"SOLLET privateKey>b58 {solKeygenWallet.Account.PrivateKey.Key}");
+            //Console.WriteLine($"SOLLET privateKey>b58 {solKeygenWallet.Account.PrivateKey.Key}");
 
-            if (solKeygenWallet.Account.PublicKey.Key != expectedSolKeygenPublicKey || solKeygenWallet.Account.PrivateKey.Key != expectedSolKeygenPrivateKey)
-            {
-                Console.WriteLine("NOT GOOD FOR THE SOL");
-            }
-            else
-            {
-                Console.WriteLine("GOOD FOR THE SOL");
-            }
+            //if (solKeygenWallet.Account.PublicKey.Key != expectedSolKeygenPublicKey || solKeygenWallet.Account.PrivateKey.Key != expectedSolKeygenPrivateKey)
+            //{
+            //    Console.WriteLine("NOT GOOD FOR THE SOL");
+            //}
+            //else
+            //{
+            //    Console.WriteLine("GOOD FOR THE SOL");
+            //}
 
         }
     }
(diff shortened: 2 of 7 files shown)


BifrostTitan:
Private key will always be exposed regardless if its a string type. The hacker would target 64 byte arrays instead of strings in the memory. Changing it to bytes does not protect anything it only makes it hard for the developer to use the class. 

Also generating new accounts would cause the private key to become irretrievable after creation. Mnemonic phrase used for wallets is stored as a string as well. Even if you wrapped the two fields in encryption the hacker would just inject code into the DLL to load up their specific class with your keys rather than the default Account class making all these changes redundant.

How do we protect the code?

- Minimize Account object exposure in code by immediately disposing of it after signing.
 No matter how you code it. The bytes will be visible in memory. Reduce the exposure time to prevent memory snooping via pointers. Meaning the hacker would have to persistently scan the memory over and over again in order to find the data instead of going right to it via pointers. Can not steal what is not there.
 
- Sign your DLLs/software and make sure the software verifies these hashes before loading private keys into a solnet DLL.

- Use Microsoft Defenders memory integrity system for your server, PC and files to prevent memory tampering on running apps. 

Just about any anti-virus software will flag & isolate software if its capable of memory snooping. This is not a real threat unless you run around with your firewall off and no protection. In this case your whole machine is compromised if they have access and can run memory dumps on your apps etc 

Here is an example of using device based encryption and memory hygiene in a desktop wallet application while still using the default solnet code.
https://github.com/Bifrost-Technologies/Tranquility/blob/master/Tranquility%20Wallet/Wallet/Wallet.cs

As long as you have signed software and encrypted external private keys you will be safe from other apps on the OS.

ManOguaR:
in my humble opinion:
The hacker would target 64 byte arrays instead of strings in the memory. Changing it to bytes does not protect anything it only makes it hard for the developer to use the class.

the difference between cases resides in using the heap or the stack, by using the heap, the key is travelling around in the .net execution system and garbage collectors, by using the stack it's volatile and not depends on garbage collectors. so harder to track...

you are right recomending secure the builds and the environment, but... code must be as safer as possible, and to keep a DebugOutput, trackable via any mstooling, easy to find, and prone to appear in any log with the key... using third party tooling... and living in the heap... ^^ well done.
    [DebuggerDisplay("Key = {ToString()}")]



________________________________
De: Nathan Martell
Enviado: Viernes, 18 de Abril de 2025 19:40
Para: bmresearch/Solnet
CC: Eduardo Ridao; Author
Asunto: Re: [bmresearch/Solnet] Refactor PrivateKey class to enforce immutability and improve key security (PR #493)

[https://avatars.githubusercontent.com/u/24855008?s=20&v=4]BifrostTitan left a comment (bmresearch/Solnet#493)<https://github.com/bmresearch/Solnet/pull/493#issuecomment-2815909184>

Private key will always be exposed regardless if its a string type. The hacker would target 64 byte arrays instead of strings in the memory. Changing it to bytes does not protect anything it only makes it hard for the developer to use the class.

Also generating new accounts would cause the private key to become irretrievable after creation. Mnemonic phrase used for wallets is stored as a string as well. Even if you wrapped the two fields in encryption the hacker would just inject code into the DLL to load up their specific class with your keys rather than the default Account class making all these changes redundant.

How do we protect the code?

  *   Minimize Account object exposure in code by immediately disposing of it after signing.
No matter how you code it. The bytes will be visible in memory. Reduce the exposure time to prevent memory snooping via pointers. Meaning the hacker would have to persistently scan the memory over and over again in order to find the data instead of going right to it via pointers. Can not steal what is not there.

  *   Sign your DLLs/software and make sure the software verifies these hashes before loading private keys into a solnet DLL.

  *   Use Microsoft Defenders memory integrity system for your server, PC and files to prevent memory tampering on running apps.

Just about any anti-virus software will flag & isolate software if its capable of memory snooping. This is not a real threat unless you run around with your firewall off and no protection. In this case your whole machine is compromised if they have access and can run memory dumps on your apps etc

Here is an example of using device based encryption and memory hygiene in a desktop wallet application while still using the default solnet code.
https://github.com/Bifrost-Technologies/Tranquility/blob/master/Tranquility%20Wallet/Wallet/Wallet.cs

As long as you have signed software and encrypted external private keys you will be safe from other apps on the OS.

BifrostTitan:
[DebuggerDisplay("Key = {ToString()}")] is ignored by the compiler on release builds. These tags will not exist in a shipped product if packaged correctly.

Heap variables(references) can be declared null to make them unreachable. The garbage collector will eventually reclaim the memory occupied by the unreachable object when it runs its cleanup process. The goal is to prevent exposure whether that is stack/heap memory.

PR #509, opened 2026-07-29
Modernize code style with C# 12 syntax

kingsznhone:
>  

| Ready | Refactor | No |  |

## Problem

The repo contains legacy coding styles left over from many years ago.

## Solution

Modernized the transaction decoding example using C# 12 syntax and improved code formatting. Replaced the traditional signature array initialization with a collection expression, improved the readability of large numeric literals, and reformatted fluent transaction-builder calls and long output statements. The transaction construction, signing, serialization, decoding, and simulation behavior remain unchanged.


BifrostTitan:
To maintain backwards compatibility with .NET standard 2.1 it has to remain the same. Software like Unity use c# 9.0 and the performance gains are minimal.

I appreciate the effort to modernize the sdk 🦾

